PR #5, opened 2015-06-17
deduping and formatting README

shalako:


atulkc:
Initial Setup header was removed but there is still a link to it in "Developer Workflow" section. That needs to be removed.

shalako:
Good catch. I updated the branch to remove that link.

PR #8, opened 2015-07-01
consolidate instructions for manual testing of diego with instructions for routing api

shalako:


atulkc:
Thanks @shalako, we have merged the changes.

PR #25, opened 2016-03-02
corrected manifest generation filename, `properties` ambiguity

dkoper:


crhino:
Finished in story above.

PR #31, opened 2016-06-09
Remove hard-coded references to *.service.cf.internal and use properties

mook-as:
Hi there!

This makes it possible to run in configurations where a different service discovery mechanism is involved.  Also cleaned up a couple instances of hard-coded port numbers for routing-api (though it isn't yet possible to make it listen on a different port).

See also PRs where we've done the same for other BOSH releases:
- cloudfoundry/cf-release#915
- cloudfoundry-incubator/etcd-release#9
- cloudfoundry-incubator/diego-release#153

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -126,12 +126,18 @@ properties:
 
   uaa.clients.gorouter.secret:
     description: "Password for UAA client for the gorouter."
+  uaa.token_endpoint:
+    description: "UAA token endpoint host name"
+    default: uaa.service.cf.internal
   uaa.port:
     description: "Port on which UAA is running."
     default: 8080
   uaa.ssl.port:
     description: "Secure Port on which UAA is running."
 
+  routing_api.uri:
+    description: "URL where the routing API can be reached internally"
+    default: http://routing-api.service.cf.internal
   routing_api.port:
     description: "Port on which Routing API is running."
     default: 3000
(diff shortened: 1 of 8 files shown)


aarondl:
Any idea on when this might be considered for merging?

shalako:
Hello @aarondl 
What are you using instead of UAA and Routing API?

aarondl:
Hey @shalako. We use UAA and Routing API. We use Fissile to Dockerize BOSH deployments and service discovery simply doesn't work with hardcoded values in that environment.

These hardcoded hostnames make assumptions about the service discovery mechanism and sort of couple to an invisible contract because only one deployment mechanism is usually used for Cloud Foundry. Making hostnames configurable is a good idea in any case, but in particular it supports different deployment strategies.

shalako:
We prefer not to make things configurable just because we can, or to imagine use cases. We prefer to keep configuration and documentation as simple as possible, until there is current and valid demand for change. There are too many things to do and our team is small, so prioritization must be brutal.

This change seems reasonable, but I'm still interested to learn how you handle service discovery in your environment.

Thank you!

aarondl:
Sorry, I thought the mention of Docker had explained it. Docker provides DNS and we use that to deal with service discovery.

shalako:
Prioritized for engineering review.

aaronshurley:
@aarondl : Closing this PR as the corresponding tracker story is completed.

Regards
Aaron & @shashwathi

PR #35, opened 2016-06-28
Update ifrit for gorouter

saliceti:
## What

gorouter relies on [ifrit](https://github.com/tedsuo/ifrit) to manage its processes. But the version pakaged in routing-release doesn't allow gorouter to receive more than one signal.
This breaks the drain script as it repeatedly sends signal USR1 while gorouter is draining the connections, until gorouter finally stops. Without the fix, gorouter dies after 5 seconds only when the drain script sends the second signal.
[This commit in ifrit](https://github.com/tedsuo/ifrit/commit/a335a27acf022a307e73546a78e31b57a7438a1f) allows to receive more than one signal. This PR updates the ifrit submodule to include this change.

## How to test

* Start gorouter with `drain_wait: 60` for example
* Run the drain script once
* Run the drain script after 5 seconds (same as bosh-agent does)
* In current version, gorouter dies. With the fix, it continues draining the connections and stops after 60s


dcarley:
We discovered this problem when implementing availability tests in our CF deployment pipeline. They make a requests every 100ms to an app that is deployed on CF whilst a `bosh deploy` of CF is happening. We were seeing several requests dropped, which coincided with the router jobs updating. With the change in this PR no requests were dropped and we were able to raise the threshold of our test to 100%.

There's some more detail about the test here if you're interested: alphagov/paas-cf#321

markstgodard:
Thanks for the PR!

We tested and merged. thanks!

@markstgodard && @shashwathi

PR #38, opened 2016-06-29
fixed path to deployment manifest

dkoper:
There are no releases/cf-routing-<lastest_version>.yml files. I assume it should be releases/routing-<lastest_version>.yml?


shalako:
Thank you, Dies. Good catch as always. We recently renamed the BOSH release from `cf-routing-release` to `routing-release` and missed this reference in our cleanup.

PR #54, opened 2016-11-08
Make tcp-router heathcheck ip and routing-api listen ip configurable

mdelillo:
This allows haproxy and routing-api to bind to specific interfaces (i.e. localhost) instead of all interfaces.

Corresponding routing-api PR: https://github.com/cloudfoundry-incubator/routing-api/pull/9

@mdelillo @aemengo

### DIFF
--- a/jobs/haproxy/spec
+++ b/jobs/haproxy/spec
@@ -21,3 +21,7 @@ properties:
   haproxy.health_check_port:
     description: "Load balancer in front of TCP Routers should be configured to check the health of TCP Router instances by establishing a TCP connection on this port"
     default: 80
+
+  haproxy.health_check_ip:
+    description: "IP for haproxy health check"
+    default: ""
--- a/jobs/routing-api/spec
+++ b/jobs/routing-api/spec
@@ -108,3 +108,7 @@ properties:
   routing_api.lock_retry_interval:
     description: "interval to wait before retrying a failed lock acquisition"
     default: "5s"
+
+  routing_api.listen_ip:
+    description: "IP for routing api to listen on"
+    default: "0.0.0.0"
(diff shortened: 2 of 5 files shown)


shalako:
Hello @mdelillo 

Why is this necessary? Please share your use cases.

mdelillo:
Hey @shalako, this is something necessary for PCF Dev. We are only exposing necessary services (such as gorouter) outside of our VM for security reasons. This will also be useful for future work on small-footprint CF.

jobs/routing-api/spec
```diff
@@ -108,3 +108,7 @@ properties:
   routing_api.lock_retry_interval:
     description: "interval to wait before retrying a failed lock acquisition"
     default: "5s"
+
+  routing_api.listen_ip:
+    description: "IP for routing api to listen on"
+    default: "0.0.0.0"
```
shalako: Why is the default for this property different from the default for `haproxy.health_check_ip`?

jobs/haproxy/spec
```diff
@@ -21,3 +21,7 @@ properties:
   haproxy.health_check_port:
     description: "Load balancer in front of TCP Routers should be configured to check the health of TCP Router instances by establishing a TCP connection on this port"
     default: 80
+
+  haproxy.health_check_ip:
+    description: "IP for haproxy health check"
```
shalako: "Interface on which TCP Router listens for healthchecks"

jobs/haproxy/spec
```diff
@@ -21,3 +21,7 @@ properties:
...
+    default: ""
```
shalako: Shouldn't this be "0.0.0.0"?

mdelillo:
Closing because of lack of consistency between this PR and the gorouter status port configuration.

PR #59, opened 2017-01-05
consume NATS properties via BOSH link if available

JamesClonk:
Signed-off-by: Fabio Berchtold <fabio.berchtold@swisscom.com>

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -19,6 +19,11 @@ packages:
   - routing_utils
   - gorouter
 
+consumes:
+- name: nats
+  type: nats
+  optional: true
+
 properties:
   router.port:
     description: "Listening Port for Router."
@@ -121,13 +126,19 @@ properties:
     default: false
 
   nats.user:
-    description:
+    description: User name for NATS authentication
+    example: nats
   nats.password:
-    description:
+    description: Password for NATS authentication
+    example: natSpa55w0rd
   nats.port:
-    description:
+    description: TCP port of NATS servers
+    example: 4222
   nats.machines:
-    description: "IP of each NATS cluster member."
+    description: IPs of each NATS cluster member
+    example: |
+      - 192.168.50.123
+      - 192.168.52.123
   router.offset:
     description:
     default: 0
--- a/jobs/gorouter/templates/gorouter.yml.erb
+++ b/jobs/gorouter/templates/gorouter.yml.erb
@@ -13,42 +13,28 @@ status:
   if_p('nats.machines') do |ips|
     nats_machines = ips.compact
   end.else do
-    if_link('nats') do
-      nats_machines = link('nats').instances.map { |instance| instance.address }
-    end
+    nats_machines = link('nats').instances.map { |instance| instance.address }
   end
 
   nats_port = nil
   if_p('nats.port') do |prop|
     nats_port = prop
   end.else do
-    if_link('nats') do
-      link('nats').if_p("nats.port") do |prop|
-        nats_port = prop
-      end
-    end
+    nats_port = link('nats').p("nats.port")
   end
 
   nats_user = nil
   if_p('nats.user') do |prop|
     nats_user = prop
   end.else do
-    if_link('nats') do
-      link('nats').if_p("nats.user") do |prop|
-        nats_user = prop
-      end
-    end
+    nats_user = link('nats').p("nats.user")
   end
 
   nats_password = nil
   if_p('nats.password') do |prop|
     nats_password = prop
   end.else do
-    if_link('nats') do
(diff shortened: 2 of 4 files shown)


shashwathi:
Hi @JamesClonk 

Thank you for submitting the PR. I can see from the commit that only route registrar job has been changed to consume NATs link and not gorouter. Any particular reason? 

Regards
Shash

JamesClonk:
Hi @shashwathi 
I was mainly interested in the route-registrar since we are using that colocated in various other bosh releases to register routes to CF. 
But you're right, it makes sense to also allow the Gorouter to consume the NATS link.
I added the changes for the gorouter too.

Regards
Fabio

shashwathi:
Hi @JamesClonk 

Thank you for updating your PR. 

Regards
Shash

JamesClonk:
When consuming a BOSH link there must only be 1 instance_group providing it under the same name, otherwise BOSH cannot decide which one to pick. 
It seems in that bosh-lite manifest used there was the "old" way of setting up availability zones by defining them in separate instances_groups/jobs, rather than using the native AZ feature. Therefore you end up with 2x (or more) NATS jobs, all of them providing the "nats" link under the name "nats".

The solution to this is to either just have 1 NATS instance_group, like for example in the big cf-deployment manifest example: https://github.com/cloudfoundry/cf-deployment/blob/e90128d6af980c575156dede7028f99bc00fb9f1/cf-deployment.yml#L68-L71

Or to explicitly name the links these instance_groups provide.
For example:
```yaml
instance_groups:
- name: nats_z1
  jobs:
  - name: nats
    release: nats
    provides:
      nats:
        as: nats_z1
    properties: ...
- name: nats_z2
  jobs:
  - name: nats
    release: nats
    provides:
      nats:
        as: nats_z2
    properties: ...
```

when consuming you can then also specify the link name explicitly:
```yaml
instance_groups:
- name: gorouter
  jobs:
  - name: gorouter
    release: routing
    consumes:
      nats:
        from: nats_z1
    properties: ...
```

there's more info on BOSH links in the docs: http://bosh.io/docs/links.html

Amit-PivotalLabs:
@JamesClonk thanks for this PR!

@shashwathi this is a known side-effect of using links in a "BOSH 1.0" manifest, and has a standard solution: https://github.com/cloudfoundry/cf-release/blob/7bb139ef2aca118764929ef78d890260f6d183b5/templates/cf.yml#L1453-L1454.  A PR to cf-release in anticipation of this change would be required.  It would be better to test this against cf-deployment.

@JamesClonk comparing the logic in this PR to the logic in the nats-release (https://github.com/cloudfoundry/nats-release/blob/32490a93bd4f2269cc0a74f77785e08bcd7f3c1d/jobs/smoke-tests/templates/config.json.erb), it appears your code will allow things like `nats_password` to simply be nil if it isn't provided as an explicit property or via a link.  If these nats properties are required (as they are in the case of route registrar), it would be better to fail fast than to allow this value to be nil.  WDYT?

/cc @dsabeti @JesseTAlford @dsboulder

JamesClonk:
@Amit-PivotalLabs ah yes, makes sense. Thanks! I changed the PR.

shashwathi:
@JamesClonk : Thank you for updating PR.
@Amit-PivotalLabs : I realised about this work around from @dsabeti. Routing team will take care of updating templates in cf-release to support this PR.

Regards
Shash

wendorf:
@JamesClonk @shashwathi I see that this PR is preferring properties over links if both are provided. Since links are the more modern approach, I believe that we should prefer them over properties. cf-mysql-release has a good example of this [here](https://github.com/cloudfoundry/cf-mysql-release/blob/a3cdac2e5b2c1fa0876019397d11bb7ac639a343/jobs/cf-mysql-broker/templates/mylogin.cnf.erb). This allows us to treat properties as fallbacks, eventually deprecating and removing them. If an operator still wishes to manually provide connection information, they can do so with [manual linking](https://bosh.io/docs/links-manual.html).

wendorf:
I just spoke with @robdimsdale and he brought up a point about backwards-compatibility with directors that do not support links. If routing-release wants to support these older directors (earlier than v255.5, released March 2016), either keeping the priority order as is would work, or checking for `respond_to?(:if_link)` as a "version check" would work.

shashwathi:
@wendorf @Amit-PivotalLabs
 
According to the above comments, it sounds like the PR looks good for now but something to note in the future to update. Is that correct? 

Regards
Shash

wendorf:
@shashwathi I did a little more talking with different people for recommendations, and it sounds like I was mistaken. It's easier (and has cleaner backward compatibility) to prefer links, so this PR looks great from my perspective.

PR #62, opened 2017-01-24
Add support for per-instance route registration.

robdimsdale:
- This enables routing to a specific instance where round-robin routing
  across each instance of an instance group is not desired.
  A concrete example of this is the proxy in cf-mysql-release, where
  each proxy has a unique HTTP dashboard and so round-robin routing
  across them would return misleading and unhelpful results.

- When a route has the optional 'prepend_instance_index' boolean set to true,
  all URIs for this route will have the instance group index prepended
  to it.
  For example, if the URIs are:
  - foo.system-domain.com
  - bar.system-domain.com
  and the instances have indicies 0 and 2,
  then the URIs registered by the first instance will be:
  - 0-foo.system-domain.com
  - 0-bar.system-domain.com
  and the URIs registered by the second instance will be:
  - 2-foo.system-domain.com
  - 2-bar.system-domain.com
  Neither instance will register the original routes:
  - foo.system-domain.com
  - bar.system-domain.com

[#135377279]

Signed-off-by: Derek Richard <drichard@pivotal.io>

### DIFF
--- a/jobs/route_registrar/spec
+++ b/jobs/route_registrar/spec
@@ -42,7 +42,7 @@ properties:
         routing updates. It must be a time duration represented as a string
         (e.g. "10s").
         It must parse to a positive time duration i.e. "-5s" is not permitted.
-      * Additionally, the 'tags' and 'health_check' keys are optional.
+      * Additionally, the 'tags', 'health_check', and 'prepend_instance_index' keys are optional.
       * 'uris' is an array of URIs to register for the 'port'.
       * 'tags' are included in metrics that gorouter emits to support filtering.
       * 'health_check' is a hash which should have 'name' and 'script_path'.
@@ -57,6 +57,13 @@ properties:
       * if the healthcheck script exits with error, the route is unregistered.
       * if a timeout is configured, the healthcheck script must exit within the timeout,
         otherwise it is terminated (with `SIGKILL`) and the route is unregistered.
+      * 'prepend_instance_index' is a boolean. When set to true the values in 'uris'
+        will be prepended with the instance index.
+        e.g. 'some-uri.system-domain.com' will become '0-some-uri.system-domain.com' on the instance
+        with index 0, and '2-some-url.system-domain.com' on the instance with index 2.
+        When this value is enabled, each instance will register its own, unique, set of uris.
+        To additionally continue to register these original uris, create another route
+        with the same uris and set 'prepend_instance_index' to false (or omit the key entirely).
     example: |
       - name: my-service
         registration_interval: 20s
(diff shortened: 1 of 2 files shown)


robdimsdale:
We also refactored the config template to be cleaner and more resilient to e.g. encoding issues (formatting the entire object as JSON is safer and nicer than building yaml by hand).

shalako:
prioritized for eng review

robdimsdale:
We rebased this PR against the latest changes; it is functionally identically. The extra commits appearing above are a consequence of rebasing

shashwathi:
@robdimsdale : We noticed on commit `nats_machines` property is renamed to `nats_ips` and the manifest property `nats.ips` is replaced by mistake.

We will fix this ourselves just wanted to let you know.

Regards
@Nino-K & Shash

shashwathi:
@robdimsdale : We realized a use case where per instance routing should not be allowed. If a component is trying to register root domain like `*.mysql.domain.com` adding prefix makes this route invalid. 
For now, we will go ahead and merge this change with an additional warning in description but @shalako has added a story to validate the URL on route registrar side.(https://www.pivotaltracker.com/story/show/138860479)

Regards
@Nino-K && Shash

robdimsdale:
@shashwathi thanks for catching `p('nats.machines')` accidentally becoming `p('nats.ips')`. We only intended to rename the local variable, and that was only for clarity, as `nats.machines` is a confusing name.

However, given that's the property in the job spec, I think it makes sense to revert that rename, keeping `nats.machines`.

Also, good observation about registering wildcards - I didn't even know that was possible with the route registrar, otherwise we would have added a warning in the description.

shashwathi:
@robdimsdale : We have a added a warning in route_registrar job [spec](https://github.com/cloudfoundry-incubator/routing-release/blob/develop/jobs/route_registrar/spec#L67-L68) about wildcard routes, if you think the description is not clear let us know and we can update it.

Regards
Shash

robdimsdale:
@shashwathi the description looks good, thanks.

shalako:
> We realized a use case where per instance routing should not be allowed. If a component is trying to register root domain like *.mysql.domain.com adding prefix makes this route invalid.

In testing we discovered this still works. When `prepend_instance_index: true`, a uri such as `*.mysql.domain.com` will be registered as `0-*.mysql.domain.com`. Requests such as `0-*.mysql.domain.com` and `0-foo.mysql.domain.com` will succeed.

shashwathi:
Closing this issue since the PR is merged

Regards
Shash

PR #71, opened 2017-03-31
Provide gorouter link

wendorf:
Useful for UAA, haproxy, and maybe more

Signed-off-by: Dan Wendorf <dwendorf@pivotal.io>

* A short explanation of the proposed change:
gorouter job now provides a link so other jobs can get its addresses (IPs or BOSH DNS)

* An explanation of the use cases your change solves
UAA and haproxy in CF release use the router IPs to register where traffic may be coming from. By getting this information from a link, the deployment manifest is easier to author.

* Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics)
Have a dummy bosh job that consumes the link and writes it to a template. See [bosh links](https://bosh.io/docs/links.html) for more information.

* Expected result after the change
The job's addresses are accessible via link

* Current result before the change
The job's addresses must be explicitly provided in jobs' properties

* [x] I have viewed signed and have submitted the Contributor License Agreement

* [X] I have made this pull request to the `develop` branch

* [X] I have run all the unit tests using `scripts/run-unit-tests`

* [X] I have run Routing Acceptance Tests and Routing Smoke Tests on bosh lite

* [X] I have run CF Acceptance Tests on bosh lite

PR #74, opened 2017-04-18
Expose route_services_secret in http-router link

wendorf:
Other jobs in the deployment that use the link will want to have the
secret provided as a link, which will help move to the credential to
CredHub and reduce duplicate operator configuration.

  Add `router.route_services_secret` to the `http-router` link.

  Other jobs in the deployment that need the secret want to be able to consume the secret as a link so the credential can be eventually managed by CredHub and to reduce duplicate operator configuration.

  Use the attached router-link-test-release.zip, which is used like silly-router.zip from #71, but adds a new line for the secret in `file.txt`.

  Releases that consume the `http-router` link can get the `route_services_secret`

  The `route_services_secret` is not accessible via a link

  #71

* [X] I have run all the unit tests using `scripts/run-unit-tests`

* [X] I have run Routing Acceptance Tests and Routing Smoke Tests on bosh lite

* [X] I have run CF Acceptance Tests on bosh lite

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -22,6 +22,8 @@ packages:
 provides:
   - name: gorouter
     type: http-router
+    properties:
+      - router.route_services_secret
 
 consumes:
 - name: nats


wendorf:
My apologies! I didn't understand how this property was [used by other releases](https://github.com/cloudfoundry/cloud_controller_ng/blob/63a69b5aa7ac7c1a3d8002017c7506e21df27261/bosh/jobs/cloud_controller_ng/templates/cloud_controller_api.yml.erb#L144). This doesn't need to be done.

PR #82, opened 2017-07-27
gorouter: expose drain_timeout

giner:

Expose drain_timeout of gorouter configuration as router.drain_timeout

Will allow to change endpoint_timeout without affecting drain_timeout
See https://github.com/cloudfoundry/gorouter/issues/181#issuecomment-318257511

Default behavior is not changed

Not being able to change endpoint_timeout without affecting drain_timeout

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -67,6 +67,10 @@ properties:
       During this time the server will reject requests to the /health endpoint.
       This accommodates requests forwarded by a load balancer until it considers the router unhealthy.
     default: 20
+  router.drain_timeout:
+    description: |
+      Maximum delay in seconds after shut down is initiated before server stops completely.
+    default: 900
   router.healthcheck_user_agent:
     description: DEPRECATED. Use /health endpoint on port specified by status.port. User-Agent for the health check agent (usually the Load Balancer).
     example: "ELB-HealthChecker/1.0"
--- a/jobs/gorouter/templates/gorouter.yml.erb
+++ b/jobs/gorouter/templates/gorouter.yml.erb
@@ -93,6 +93,7 @@ routing_api:
 <% end %>
 
 drain_wait: <%= p("router.drain_wait") %>s
+drain_timeout: <%= p("router.drain_timeout") %>s
 healthcheck_user_agent: <%= p("router.healthcheck_user_agent") %>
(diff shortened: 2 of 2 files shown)


jobs/gorouter/spec
```diff
@@ -67,6 +67,10 @@ properties:
       During this time the server will reject requests to the /health endpoint.
       This accommodates requests forwarded by a load balancer until it considers the router unhealthy.
     default: 20
+  router.drain_timeout:
+    description: |
+      Delay in seconds after shut down is initiated before server stops completely.
```
CAFxX: `Maximum delay`?

shashwathi:
@shalako : Looks good from engg side. Feel free to prioritize this.

shalako:
What is `drain_timeout` for?
Why would someone want to change it?
You want to change it independently from `router.request_timeout_in_seconds`?

giner:
I don't need to change drain_timeout but I need to change request_timeout without changing drain_timeout.

giner:
Friendly ping.

charleshansen:
@giner looking at this now, why do you need to change drain_timeout separately from request_timeout? If the drain timeout is less than the request_timeout, you may prematurely close some connections when you roll the router.

giner:
@charleshansen, I need to have drain_timeout longer then request_timeout. To be more specific, I want to be able to keep drain_timeout 15 minutes as it is configured now but reduce request_timeout to 1-2 minutes. This way I can make routing layer more stable in certain conditions without shortening drain_timeout.

shalako:
What is the difference between request_timeout and drain_timeout in this context?
Why do you want request_timeout short and drain_timeout long?
How does this make the routing layer more stable?
I assume you understand that request_timeout governs the time allowed for app uploads?

aaronshurley:
@giner do you still wish to proceed with this PR? If so, could you please provide responses to @shalako above?

giner:
Thank you for the ping. Need some time to go through this once again. Will update soon.

giner:
I've checked this once again and I think there is no good reason to merge it. My understanding of endpoint_timeout in gorouter configuration was wrong as I thought it was a read/write timeout but apparently it is not.

aaronshurley:
@giner Thank you for following up! We're going to close this out for now.

PR #85, opened 2017-09-20
Fix section of router groups in readme

kei-yamazaki:
Currently still only one router groups are supported?
As I verified using cf-deployment, it seems that multiple router groups are supported.


kei-yamazaki:
Thank you for your response.
I misunderstood because routing-api can define multiple router groups.

PR #106, opened 2018-04-27
Add microseconds to logs

giner:
Seconds are often not enough especially when sorting is required.

Timestamp is in ISO 8601.

Before:
```
$ echo test | prepend_datetime
[2018-04-27 20:37:20+0900] test
```

After:
```
$ echo test | prepend_datetime
2018-04-27T20:37:32.618717+0900 test
```

### DIFF
--- a/src/routing_utils/syslog_utils.sh
+++ b/src/routing_utils/syslog_utils.sh
@@ -12,5 +12,5 @@ function tee_output_to_sys_log() {
 }
 
 function prepend_datetime() {
-  perl -ne 'BEGIN { use POSIX strftime; STDOUT->autoflush(1) }; my $time = strftime("[%Y-%m-%d %H:%M:%S%z]", localtime); print("$time $_")'
+  perl -ne 'BEGIN { use Time::HiRes "time"; use POSIX "strftime"; STDOUT->autoflush(1) }; my $t = time; my $fsec = sprintf ".%06d", ($t-int($t))*1000000; my $time = strftime("%Y-%m-%dT%H:%M:%S".$fsec."%z", localtime $t); print("$time $_")'
 }


shalako:
Hello @giner 
We reviewed the change with the Engineering team today and noticed that in addition to adding microseconds, you've also changed the format; adding square brackets and the `T`. We're concerned about backward compatibility for log consumers.

giner:
Hi @shalako 
As it's likely not to be parsable after the change anyway I was thinking to make it comply with ISO standard. This would allow to use standard parsers and discourage future non-standard changes. If you think that we should keep square brackets and space let's keep it.

shubhaat:
Hi @giner , 

I'd prefer to keep the existing format to not break existing consumers. We can do the format change to make it ISO compliant in another change, after we announce it, and let operators know about it.

Regards, 
Shubha

aaronshurley:
Hi @giner,

We manually merged the PR so that we could follow it up with another [commit](https://github.com/cloudfoundry/routing-release/commit/961611bb96f1df866301cbf8c3ddd20b7c199ed3) that retains the log format for now.

Thanks!

shubhaat:
Hi @giner , 

We actually merged this change, but unfortunately are currently working on using BPM (https://github.com/cloudfoundry-incubator/bpm-release) for all routing components, and actually for components across Cloud Foundry. BPM changes and standardizes the log formatting across components in CF, and the change you made to add microseconds will be lost when we cut a release. BPM provides better security for CF component VMs by enabling process isolation as described in the docs. 

To read more about the runtime components refer to [this](https://github.com/cloudfoundry-incubator/bpm-release/blob/master/docs/runtime.md.)  

The time in the log  would look like: 

```
[2018-08-06T17:02:03.840+0000]
```
Please feel free to reach back to me with questions. 

Regards, 
Shubha

PR #111, opened 2018-06-13
Config driven host for route-registrar

### LINKED ISSUE #108: Unable to pass host ip as part of config
**Context:**

We want to implement a High-Availability setup for our deployment. Earlier, we used to have a single ip for our deployment. But, with this new setup, we intend to have 2 Ips and a floating ip pointing to the master(default) or slave(failover). 

**Our Config:**
We have an external URL, which we register with the ip using the routing-release. 
In earlier setup, the `registrar_settings.json` used to look somewhat like:

```
{
"host": "10.11.252.10", 
  "routes": [
    {
      "health_check": {
        "name": "health-check",
        "script_path": "path_to_health_check_script",
        "timeout": "5s"
      },
      "name": "deployment-name",
      "port": 9292,
      "registration_interval": "20s",
      "tags": {
        "component": "component-name",
        "env": "production"
      },
      "uris": [
        "externalhost.dummy.com"
      ]
    }
  ]
}
```

**Problem:**
Now, with the HA setup, we ideally would like register the uri with the floating-ip. 
("host": floating-ip)

However, as I see from the job settings of route-registrar (https://github.com/cloudfoundry/routing-release/blob/develop/jobs/route_registrar/templates/registrar_settings.json.erb#L50), the job picks up the ip of the VM it is deployed on. In the HA setup, the routing-release will be present in both the VMs and will be attempting to register the uri with both the ips, which is undesirable.

Is there a way we can pass the host-ip as part of config? 
Do you have any suggestion on how to handle the routing in our scenario?

sauravmndl:
It fixes #108: Idea is to read host from job spec property
if not found default to spec.ip
This intriduces flexibilty as other parameters, keeping existing functionality intact.
This would help other bosh releases consuming route-registrar job from routing-release,
to have flexibility to register routes with configurable host.

### DIFF
--- a/jobs/route_registrar/spec
+++ b/jobs/route_registrar/spec
@@ -33,6 +33,10 @@ properties:
     description: Password for NATS authentication
     example: natSpa55w0rd
 
+  host:
+    description: Destination hostname or IP for the routes being registered
+    example: 192.168.60.25
+
   route_registrar.routes:
     description: |
       * Array of hashes determining which routes will be registered.
--- a/jobs/route_registrar/templates/registrar_settings.json.erb
+++ b/jobs/route_registrar/templates/registrar_settings.json.erb
@@ -45,9 +45,16 @@
     end
   end
 
+  host = nil
+  if_p('host') do |prop|
+    host = prop
+  end.else do
+    host = spec.ip
+  end
+
   config = {
     message_bus_servers: message_bus_servers,
-    host: spec.ip,
+    host: host,
     routes: routes,
   }
   JSON.pretty_generate(config)


shubhaat:
@sauravmndl - thanks for making this PR, I am curious what you intend to use this for? Do you have a component that needs the flexibility to change the flexibility to register routes with configurable host? 

Regards, 
Shubha

sauravmndl:
@shubhaat Indeed, I have a use case where, in a bosh deployment 2 vms are are behind a virtual ip. The virtual ip gets attached and detached to vms based on certain condition. I also need the virtual ip to be registered with a dns url. So, from a vm I need to register virtual ip not vm's fixed ip. Hence, need the flexibility. Also, the way I made the change in this PR, doesn't change existing functionality of routing-release provided job route-registrar. Existing consumer can continue to use the exiting routing-release or upcoming one with my code change without doing any change from their side.

aaronshurley:
@shubhaat thoughts on the above? If we choose to do this, we'd want to mention the default behavior in the property description

aaronshurley:
This was briefly discussed in [Slack](https://cloudfoundry.slack.com/archives/C05586PBX/p1530566204000212) and it sounds like we have product support. Awaiting prioritization.

shubhaat:
This is now in routing-release [0.180.0](https://github.com/cloudfoundry/routing-release/releases/tag/0.180.0)

PR #116, opened 2018-08-11
Clarify GCP TCP setting for `frontend_idle_timeout`

ljfranklin:
We recently noticed strange "connection reset by peer" occasionally when running test suites with a GCP TCP LB.  Turns out GCP will forcibly cut all idle TCP connections after 10 minutes: https://cloud.google.com/compute/docs/troubleshooting/general-tips#communicatewithinternet. With the default value of 900 seconds for `router.frontend_idle_timeout` our app would come up successfully and open a keep-alive connection through the TCP LB to the gorouter, but the first request after the 10 minute mark would result in "connection reset by peer". Looks like GCP cuts the connection without shutting down the keep-alive connection properly. So to use a GCP TCP LB you need to set `frontend_idle_timeout` to something less than 600 seconds. Setting this property to 60 seconds fixed the flakiness for us. However, for GCP **HTTP** LBs you still want to use a value over 600 seconds as described here: https://blog.percy.io/tuning-nginx-behind-google-cloud-platform-http-s-load-balancer-305982ddb340. Fun times.

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -167,8 +167,8 @@ properties:
     description: |
       (optional, integer) Duration in seconds to maintain an open connection when client supports keep-alive.
       This property must be configured with regards to how an IaaS-provided load balancer behaves in order to prevent connections from being closed prematurely.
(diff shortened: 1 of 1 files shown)


zachgersh:
@ljfranklin cheers for the clarification!

PR #117, opened 2018-08-12
Add possibility to define linux local_port_range

kayrus:

certain firewall implementation doesn't pass through the connections that was made from source ports lower than 32768. This patch allows to define a local port range and override the default "1024 65535" local port range.

* [ ] I have made this pull request to the `develop` branch

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -268,5 +268,9 @@ properties:
     default: false
 
   uaa.ca_cert:
-    description : "Certificate authority for communication between clients and uaa."
+    description: "Certificate authority for communication between clients and uaa."
     default: ""
+
+  router.ip_local_port_range:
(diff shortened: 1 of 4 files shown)


shubhaat:
Hi @kayrus - thanks for making this PR, when you say certain Firewalls, do you have a specific Firewall you are using that does not support the default port range, just curious. 

Regards, 
Shubha

kayrus:
> thanks for making this PR, when you say certain Firewalls, do you have a specific Firewall you are using that does not support the default port range, just curious.
vmware stateless DVS rules. we use 32k-65k as a workaround for them.

kayrus:
@aaronshurley I see there is a new release, however a new `router.ip_local_port_range` option is not mentioned in the release notes.

PR #126, opened 2018-10-19
Allow set HSTS header value for gorouter

keymon:
What?
-----

Add the changes necessary to use the feature added in
https://github.com/cloudfoundry/gorouter/pull/231

We want to be able to inject the HSTS header, Strict-Transport-Security
into the response of the gorouter[1]

A new option strict_transport_security_header is added and can be
set to any free string to be inserted in this header, the
operator must set the right format.

The header would be injected only if configured and if the backend
does not return that header.

Use case
--------

In our organisation there is a requirement of any application to use HTTPS and set HSTS headers. We want every application to respond with these headers. The tenant can always override them if they need to.

How to review/test
----------------------

 - Set the right config `strict_transport_security_header` with the right value. Check that gorouter returns the header as desired.

Related PR
----------------
https://github.com/cloudfoundry/gorouter/pull/231 should be merged and the commit referencing the submodule updated or deleted.

Checklist
---------

* [x] I have run all the unit tests using `scripts/run-unit-tests`

* [x] I have run Routing Acceptance Tests and Routing Smoke Tests on bosh lite

* [x] I have run CF Acceptance Tests on bosh lite

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -167,6 +167,9 @@ properties:
       When force_forwarded_proto_https: true, this property will be ignored.
       Otherwise,  we recommend setting the property to true if Gorouter is the first component to terminate TLS, and set to false when your load balancer is terminating TLS and setting the X-Forwarded-Proto header.
     default: false
+  router.strict_transport_security_header:
+    description: (optional, string) If set, gorouter will inject the HSTS (HTTP Strict Transport Security) Header if not set by the backend into the response using the provided value. The value must follow the HSTS spec.
+    example: "max-age=31536000; includeSubDomains; preload"
   router.frontend_idle_timeout:
     description: |
       (optional, integer) Duration in seconds to maintain an open connection when client supports keep-alive.
--- a/jobs/gorouter/templates/gorouter.yml.erb
+++ b/jobs/gorouter/templates/gorouter.yml.erb
@@ -179,6 +179,10 @@ token_fetcher_expiration_buffer_time: 30
 enable_proxy: <%= p("router.enable_proxy") %>
 force_forwarded_proto_https: <%= p("router.force_forwarded_proto_https") %>
 sanitize_forwarded_proto: <%= p("router.sanitize_forwarded_proto") %>
+<% if_p("router.strict_transport_security_header") do |h| %>
+strict_transport_security_header: <%= h %>
+<% end %>
+
 pid_file: /var/vcap/sys/run/gorouter/gorouter.pid
 
 ip_local_port_range: <%= p("router.ip_local_port_range") %>


rosenhouse:
hey @keymon ,

We're looking at making a slightly more generic interface for this...

What do you think about the BOSH interface illustrated in [this tracker story](https://www.pivotaltracker.com/story/show/161518054)?

cc @shalako @shubhaat 

Thanks!

keymon:
> We're looking at making a slightly more generic interface for this...

For me great, i did it this way because I was told not to do it generic in slack. I think it will be really easy to implement it as you say, I'll do a different pr later

keymon:
Some comments are:

- might want to have multiple headers with same name
- default behaviour right now is: inject if missing, but in the future we might want to have other logic: override, append, delete

Because that, I suggest the config is a list of hash with name of the header and value

keymon:
Some implementation of previous idea in : https://github.com/cloudfoundry/routing-release/pull/128 and https://github.com/cloudfoundry/gorouter/pull/232

shubhaat:
Hi @keymon , 

This looks great, we want to clarify the use case for having both `inject_response_headers` and `append_response_headers`. Are there specific use cases you have in mind for `inject_response_headers`? 

Shubha

keymon:
> This looks great, we want to clarify the use case for having both inject_response_headers and append_response_headers. Are there specific use cases you have in mind for inject_response_headers?

Actually, there can be multiple semantings.
 -  `inject_response_headers` == `add_response_headers_unless_set`: Adds the headers if and only if the proxied backend (or handlers) did not set them already. 
     
     That would allow apps to override the headers, which is what we want for HSTS headers in our use case.
      I picked the  `inject` verb, but it is not really that... we can use `add_response_headers_unless_set` or whatever. Open to suggestions,
 - `set_response_headers`: Adds the headers, overriding the headers with same name set by the app.
 - `append_resppnse_headers`: Would append the headers. If the app sets headers with the same name, there will be the ones set by the app AND the ones set by the router (after)
 - `delete_response_headers`: deletes them

> Have you implemented inject, append and delete in your PRs?

I only implemented the `inject` (aka `add_unless_set`). If you are happy with my implementation, add other operations would be very easy. I can do it in the same PR or later.

shubhaat:
Thanks @keymon we will prioritize the stories for our team to review the PR and merge it, and we have another story for deleting headers as well. We will prioritize delete headers once we merge this, and are happy to get a PR too.

shalako:
In a previous iteration of your comments, I saw a property structure that used `action`. That looked pretty good; was there a reason to decided against it?
```
http_rewrite:
  responses:
    headers:
    - name: foo
      value: bar
      action: {strip_and_set, set_only_if_absent, always_append, delete}
```

keymon:
> In a previous iteration of your comments, I saw a property structure that used action. That looked pretty good; was there a reason to decided against it?

I was thinking to:
 - be able to easily modify that using opsfile
 - leverage the release spec to check the operations (e.g. if I type wrongly `set_if_absent` rather than `set_only_if_absent`

But happy to implement it in that way. We can do some checking in the gorouter, and blow up if the action is unknown or so.

zachgersh:
Hi @keymon, we are going to close this because it seems like #128 is the more relevant PR, and this one corresponds to the already closed PR in gorouter.

PR #199, opened 2021-03-18
add sni routing support

pnikonowicz:
Adds resource lock key to routing-api. (#3)

added sni fields in readme

adds the following fields for sni routing:
add sni_port
add sni type

this type will eventually get adapated back into a tcp type,
but explictly adding an sni type allows for a more clear
api for the user.

* [ ] I have made this pull request to the `develop` branch

* [ ] I have run all the unit tests using `scripts/run-unit-tests-in-docker`

### DIFF
--- a/jobs/route_registrar/spec
+++ b/jobs/route_registrar/spec
@@ -93,9 +93,10 @@ properties:
 
       route object
         name (required, string, for all routes): Human-readable reference for the route
-        type (optional, string, for all routes): Defaults to http, can specify http or tcp.
+        type (optional, string, for all routes): Defaults to http, can specify http, sni, or tcp.
         uris (required, array, for http routes): When Gorouter receives a request that matches one of these URIs,
           it will forward them to the IP of the host on which route_registrar runs, and either port or tls_port.
+        sni_port (required, integer, for sni rotues): When sni type provided, this is the downstream port to route to
         port (required, integer, for all routes): Either `port` or `tls_port` are required; if both are provided, Gorouter will prefer tls_port.
           Requests for associated URIs will be forwarded unencypted by the router to this port.
           The IP is determined automatically from the host on which route-registrar is run.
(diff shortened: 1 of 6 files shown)


goonzoid:
Thanks @pnikonowicz. I'm curious: what's the use case for exposing `routing_api.lock_resource_key`. When would an operator need to set that to something other than the default?

goonzoid:
Additionally, I don't think this needs to be a blocker for merging this, but do you think you could help us figure out a sensible example to add for an `sni` route in the example block below `route_registrar.routes` in `jobs/route_registrar/spec`? I'm not super clear on how `server_cert_domain_name_modifier` is likely to be used.

pnikonowicz:
> Thanks @pnikonowicz. I'm curious: what's the use case for exposing `routing_api.lock_resource_key`. When would an operator need to set that to something other than the default?

This is essentially an "extract parameter" refactor that we did in order to have 2 routing apis running in production (which we currently have in our 1.13 release). 

We could have removed this change as it will probably not be needed by anyone going forward, but we liked the flexibility that this change brings and thought it to be a nice addition to the code base. This is of course, an opinion and not a logical certainty. 

In other words, it was useful to us in the past, maybe it will be useful to someone else in the future. Also, we know the feature works as we test it in our CI pipelines.

pnikonowicz:
> Additionally, I don't think this needs to be a blocker for merging this, but do you think you could help us figure out a sensible example to add for an `sni` route in the example block below `route_registrar.routes` in `jobs/route_registrar/spec`? I'm not super clear on how `server_cert_domain_name_modifier` is likely to be used.

It's difficult to explain `server_cert_domain_name_modifier`. The need for it comes from our specific domain. It essentially just swaps one name for another. 

Some additional context as to why it exists, we have some configurations that allow 2 services to be on 1 vm. Those 2 services must be routable by specific names and cannot be used by the vm hostname directly as the port information is lost when going through the SNI proxy. 

Another way of saying it, service `A` on port `1` and service `B` on port `2` running on the same VM  need a unique name to be routable. So if the hostname for the VM is `vm.info` we cannot have 2 routes to `vm.info`, so we substitute the hostname using `server_cert_domain_name_modifier` in order to make the hostnames unique.

PR #201, opened 2021-04-01
Router Router for extra routing

0401-7r011:

**No no, thank you**

**Once upon a time, there lived a kingdom of routing, and in that routing kingdom... wait wait, let me try that again. You see, routing, is a lot like parfait. Both are delicious, and both have layers. And when you're done with one, you want more. So this change takes a great thing that is routing, and adds more layers, which makes it greater, just like parfait.**

**Rather than traffic coming in and going through one routing layer. This change lets you deploy layers of gorouters. Each layer routing traffic to the next. Only the final delicious layer actually listens to NATS updates and routes data to the applications. The other layers are configured to route to each other using bosh links, or to the final routing layer I suppose.**

**Packets reaching applications too quickly?**

**There is a [separate pull request](https://github.com/cloudfoundry/gorouter/pull/277) in the gorouter that makes the needed changes there. At that point you can just add a new routing layer to your deployment manifest, configure the links appropriately (this is important since you are deploying two copies of the router job, and they both produce and consume the same link, you wouldn't want a layer to route to itself, that would be inefficient).**
```yaml
instance_groups:
  ...
  - name: router_router
    jobs:
      name: gorouter
      consumes:
        router_router_target_router: from: router_router_target}
      properties:
        ...
        router_router: true
    vm_extensions:
      - extension-for-lb
  - name: router
    jobs:
      name: gorouter
      consumes:
        router_router_target_router: nil
      provides:
        gorouter: {as: router_router_target}
      properties:
        ...
        router_router: false
```

**Layers... remember?**

**!Layers**

https://github.com/cloudfoundry/gorouter/pull/277

~~* [ ] I have viewed signed and have submitted the Contributor License Agreement~~

* [👍 ] I have made this pull request to the `develop` branch

* [👍] I have run all the unit tests using `scripts/run-unit-tests-in-docker`

~~* [ ] (Optional) I have run Routing Acceptance Tests and Routing Smoke Tests on bosh lite~~

~~* [ ] (Optional) I have run CF Acceptance Tests on bosh lite~~

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -25,6 +25,8 @@ packages:
 provides:
   - name: gorouter
     type: http-router
+    properties:
+    - router.port
 
 consumes:
 - name: nats
@@ -32,6 +34,9 @@ consumes:
   optional: true
 - name: routing_api
   type: routing_api
+  optional: true
+- name: router_router_target_router
+  type: http-router
   optional: true
 
 properties:
@@ -405,3 +410,9 @@ properties:
   router.send_http_start_stop_client_event:
     description: "Send a httpstartstopevent of type client for each request"
     default: true
+
+  router.router_router:
+    description: |
+      Ignore NATS routing updates, and route all traffic to the routers consumed
+      by the router_router_target_router bosh link
+    default: false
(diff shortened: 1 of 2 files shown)


ameowlia:
Sorry we didn't get this merged in time 🙃

PR #214, opened 2021-07-27
Remove reliance on bosh link named 'nats'

46bit:
In https://github.com/cloudfoundry/cf-deployment/issues/929 we plan to remove plaintext NATS, and enforce that everything uses TLS NATS. The easiest way to do this is to remove the `nats` BOSH link, and have only the `nats-tls` BOSH link.

This commit changes a file that required the `nats` BOSH link to exist. The file will now only use the `nats-tls` BOSH link if told to use TLS NATS.

### DIFF
--- a/jobs/route_registrar/templates/registrar_settings.json.erb
+++ b/jobs/route_registrar/templates/registrar_settings.json.erb
@@ -1,14 +1,19 @@
 <%=
   require 'json'
 
+  nats_link_name = 'nats'
+  if p('nats.tls.enabled')
+    nats_link_name = 'nats-tls'
+  end
+
   nats_machines = nil
   if_p('nats.machines') do |ips|
     nats_machines = ips.compact
   end.else do
-    link('nats').if_p("nats.hostname") do |hostname|
+    link(nats_link_name).if_p("nats.hostname") do |hostname|
       nats_machines = [hostname]
     end.else do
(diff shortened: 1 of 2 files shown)


46bit:
Ready for review :)

ameowlia:
Looks good 👍 I tested with both the nats-tls and nats configuration.

ameowlia (approved):

acrmp:
Hi @46bit,

Thanks for this PR! We really appreciate the work being done to move to TLS everywhere for NATS.

* We noticed that the `nats.tls.hostname` property effectively removed in this PR [is still present in the job spec](https://github.com/cloudfoundry/routing-release/blob/6b13e6092d360b7e621434258ee49a7ca3ebc4c0/jobs/route_registrar/spec#L56-L57). Would you be able to also remove the property from the job spec to avoid confusion? We'll want to point out the removal of the property in the release notes when we next cut a release.

* Thanks for also correcting the test examples while you were there to use `nats.hostname` instead of `nats.host`. I think there were still a couple of references to `nats.host` left over if you wanted to fix those too. 😄

Thanks!

Andrew.

46bit:
@acrmp @ameowlia Thanks for the feedback, I've filed https://github.com/cloudfoundry/routing-release/pull/216 to address that. If we can get this shipped in a new version soon that'd be a big help internally for SAP.

PR #238, opened 2021-10-22
Fail on cert without san

geofffranks:
Updates the gorouter template to fail when rendering, if a cert provided in the`router.tls_pem` list does not contain a SAN. As of golang 1.17, this will be a requirement, so adding these warnings to allow people to be aware of the issue. After encountering it, you can set `golang.x509ignoreCN` to avoid the error, until such a time as you get your cert upgraded, but this stopgap will cease to work after gorouter is built against golang 1.17.

### DIFF
--- a/jobs/gorouter/templates/gorouter.yml.erb
+++ b/jobs/gorouter/templates/gorouter.yml.erb
@@ -268,7 +268,18 @@ if p('router.enable_ssl')
     if !cert_pair.is_a?(Hash) || !cert_pair.key?('cert_chain') || !cert_pair.key?('private_key')
       raise 'must provide cert_chain and private_key with tls_pem'
     end
+
+    cert = OpenSSL::X509::Certificate.new cert_pair['cert_chain']
+    has_san = cert.extensions.map { |ext|
+      x509ext = OpenSSL::X509::Extension.new ext
+      x509ext.oid == "2.5.29.17" # https://oidref.com/2.5.29.17
+    }.reduce(:|)
+
+    if ! p('golang.x509ignoreCN') and ! has_san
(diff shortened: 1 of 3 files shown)


jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -268,7 +268,18 @@ if p('router.enable_ssl')
     if !cert_pair.is_a?(Hash) || !cert_pair.key?('cert_chain') || !cert_pair.key?('private_key')
       raise 'must provide cert_chain and private_key with tls_pem'
     end
+
+    cert = OpenSSL::X509::Certificate.new cert_pair['cert_chain']
+    has_san = cert.extensions.map { |ext|
+      x509ext = OpenSSL::X509::Extension.new ext
+      x509ext.oid == "2.5.29.17" # https://oidref.com/2.5.29.17
```
reneighbor, line 275: 👀 🏅

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -268,7 +268,18 @@ if p('router.enable_ssl')
...
+    }.reduce(:|)
+
+    if ! p('golang.x509ignoreCN') and ! has_san
+      raise 'tls_pem must include a SAN entry'
```
reneighbor, line 279: 👍

reneighbor:
This looks really great! 
* Really cool function for decrypting the cert and detecting the SAN. Thank you for sharing that knowledge.
* Thank you for adding tests for changes that came before

I had a bunch of product Qs that are beyond the scope of this PR so added them to the tracker story.

PR #239, opened 2021-10-25
Add tls_handshake_timeout paramter to override the default 10 seconds

ywei2017:
 Add tls_handshake_timeout so it can be configured

* An explanation of the use cases your change solves: 
When the diego_cells go through vSphere vMotion, there is a stun time (like the VM is frozen). When that happens, new gorouter connections to the diego_cell will timeout after 10 seconds. The parameter will allow us to shorten the timeout, so gorouter would attempt another diego_cell, mitigating the long wait (or 499 errors) on the clients.

Inspect /var/vcap/jobs/gorouter/config/gorouter.yml. When the parameter is not provided, it should say "tls_handshake_timeout: 10s"; or whatever value is provided by the operator.

Gorouter clients see faster response time during vMotion, or fewer 499 erros

Gorouter clients experience timeouts.

N/A

* [X] I have run all the unit tests using `scripts/run-unit-tests-in-docker`

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,8 @@ rubocop ./spec/
 
 If you do not have `rspec` or `rubocop` installed locally, run
 `./scripts/start-docker-for-testing.sh` and execute the commands in the docker
-container.
+container. Prepend "sudo" to he script if you are an unpreviledged user.
+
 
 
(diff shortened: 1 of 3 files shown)


geofffranks:
Thanks @ywei2017 !

PR #247, opened 2021-12-06
Create simple NATS client for debugging to replace broken publish_to_nats script

peterellisjones:
#### A short explanation of the proposed change:

This replaces `/var/vcap/jobs/gorouter/bin/publish_to_nats.erb` with a new binary that can both publish and subscribe to NATS to help debugging. The binary is a [simple go CLI](https://github.com/cloudfoundry/routing-release/blob/81281b2d80070c3f1f1916ebd275e2800679726c/src/routing_utils/nats_client/main.go) that uses the [golang NATS client](https://github.com/nats-io/nats.go) and code from the gorouter for parsing the gorouter config file.

#### An explanation of the use cases your change solves

With the introduction of TLS for NATS the existing `publish_to_nats` script **no longer works as it uses netcat**.

#### Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics)

1. Deploy the release
2. Publish a message eg `bosh -d cf ssh router/0 /var/vcap/jobs/gorouter/bin/nats_client publish router.register '{"host":"172.217.6.68","port":80,"uris":["bar.example.com"]}'`
3. Subscribe to messages eg `bosh -d cf ssh router/0 /var/vcap/jobs/gorouter/bin/nats_client subscribe 'router.*'`

#### Expected result after the change
It is easier to debug NATS issues 

#### Current result before the change
It is harder to debug NATS issues 

#### Links to any other associated PRs
N/A

* [ ] I have run all the unit tests using `scripts/run-unit-tests-in-docker`
  I have tested this manually by creating a release and asserting that the comands have the expected output


src/routing_utils/nats_client/main.go
```diff
@@ -0,0 +1,193 @@
...
+               Example: /var/vcap/jobs/gorouter/bin/nats_client subscribe 'router.*'
+
+  publish      Publish the provided message JSON to SUBJECT subscription. SUBJECT and MESSAGE are required
+               Example: /var/vcap/jobs/gorouter/bin/nats_client publish router.register '{"host":"172.217.6.68","port":80,"uris":["bar.example.com"]}'
+`
+
+// Simple NATS client for debugging
+// Uses gorouter.yml for config
+func main() {
+	if os.Args[len(os.Args)-1] == "--help" || os.Args[len(os.Args)-1] == "-h" || os.Args[len(os.Args)-1] == "help" {
```
domdom82, line 30: why not use flag or kingpin for proper args handling?

src/routing_utils/nats_client/main.go
```diff
@@ -0,0 +1,193 @@
...
+		fmt.Fprintf(os.Stderr, "Subscribing to %s\n", subject)
+		subscription, err := natsConn.SubscribeSync(subject)
+		if err != nil {
+			panic(err)
+		}
+
+		for {
+			message, err := subscription.NextMsg(time.Second * 60)
+			if err != nil {
+				panic(err)
```
domdom82: I'm not sure we should panic on time out. We could simply print that we timed out and leave it up to the user. It may not be a fatal situation.

domdom82 (changes requested):

src/routing_utils/nats_client/main.go
peterellisjones: Because I wanted to keep the number of dependencies small and keep the implementation as simple as possible.

I'm happy to add a args library if you prefer though

src/routing_utils/nats_client/main.go
peterellisjones: Fixed

src/routing_utils/nats_client/main.go
```diff
@@ -0,0 +1,193 @@
...
+		if err != nil {
+			panic(err)
+		}
+
+		for {
+			message, err := subscription.NextMsg(time.Second * 60)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "error fetching message: %s", err.Error())
+			} else {
+				fmt.Println(string(message.Data))
```
domdom82, line 97: Just stumbled across this while testing:
when you don't subscribe to a subject you will subscribe to `router.*` but here we don't print the subject this message was about, so we don't actually know whether the message is about a route registering or unregistering. 
Should be useful to also print the subject here.

src/routing_utils/nats_client/main.go
domdom82: We can leave it like it is for now and consider adding a flag parser once we see the need for additional arguments.

src/routing_utils/nats_client/main.go
peterellisjones: Updated :)

domdom82 (approved):

domdom82:
@ameowlia LTGM

PR #249, opened 2021-12-14
Add endpoint_dial_timeout_in_seconds to spec

domdom82:
Add spec to make EndpointDialTimeout configurable. It is currently hard-coded at 5s.

Some of our customers have apps that need to restart frequently and have thousands of clients connected. After the restart all those clients try to reconnect at once which temporarily puts the app under pressure such that the underlying OS can't keep up with accepting connections quick enough to not hit the 5s timeout. Being able to increase the timeout gives the app more time to eventually accept all connections without disruption clients with a 502.

1. Merge companion PR in [gorouter](https://github.com/cloudfoundry/gorouter/pull/302)
2. The BOSH manifest now allows to set an `endpoint_dial_timeout_in_seconds` property which defaults to 5 as before
```
(...)
  jobs:
  - name: gorouter
    properties:
      endpoint_dial_timeout_in_seconds: 10    <--- set to 10s in this example
      (...)
```
3. Deploy routing-release with new dial timeout

dial timeout is now configurable but remains at 5s by default

dial timeout is hard-coded at 5s

https://github.com/cloudfoundry/gorouter/pull/302

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -346,7 +346,7 @@ properties:
   endpoint_dial_timeout_in_seconds:
     description: |
       Maximum time in seconds for gorouter to establish a TCP connection with a backend. This timeout comes before `tls_handshake_timeout_in_seconds`
-      and `request_timeout_in_seconds`.
+      and `request_timeout_in_seconds`. This timeout also applies as a read timeout for websocket requests.
     default: 5
   tls_handshake_timeout_in_seconds:
     description: |
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -343,7 +343,11 @@ properties:
       are idle for longer than this duration will be closed. Related properties: `router.max_idle_connections`
       and `router.keep_alive_probe_interval`.
     default: 900
-
+  endpoint_dial_timeout_in_seconds:
+    description: |
+      Maximum time in seconds for gorouter to establish a TCP connection with a backend. This timeout comes before `tls_handshake_timeout_in_seconds`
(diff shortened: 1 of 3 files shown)


jobs/gorouter/spec
```diff
@@ -343,7 +343,11 @@ properties:
       are idle for longer than this duration will be closed. Related properties: `router.max_idle_connections`
       and `router.keep_alive_probe_interval`.
     default: 900
-
+  endpoint_dial_timeout_in_seconds:
+    description: |
+      Maximum time in seconds for gorouter to establish a TCP connection with a backend. This timeout comes before `tls_handshake_timeout_in_seconds`
```
stefanlay, line 348: The setting is also used to wait for read response from WebSockets: https://github.com/cloudfoundry/gorouter/blob/65207dc5dafa659d343975216701fb2e29e3992a/proxy/handler/request_handler.go#L71 -> https://github.com/cloudfoundry/gorouter/blob/main/proxy/handler/forwarder.go#L40 -> https://github.com/cloudfoundry/gorouter/blob/main/proxy/utils/response_reader.go#L30

Should we also document that?

domdom82: There's debate around changing that and making a new timeout specifically for websockets. I am planning on a another PR for that once we assessed the implications of such a change. Meanwhile we can add a statement that this also affects the websocket read timeout and then later remove it again..?

domdom82: @stefanlay I've added a statement for ws timeouts. We can remove it again in a follow-up PR.

stefanlay (approved):

PR #268, opened 2022-03-22
Fixing issue 250: Return a 503 not a 404 when all instances down

### LINKED ISSUE #250: Return a 503 not a 404 when all instances down
## Issue

When all instances of an application are not available, clients get 404 Not Found instead of 503 Service Unavailable.

## Affected Versions

PCF 2.12

## Context

If a route is registered but all instances of an application are currently down then GoRouter currently returns 404 Not Found. It would 
be better if it returned 503 Service Unavailable.

It looks like this issue was already discussed in the past:
https://github.com/cloudfoundry/gorouter/pull/251
https://github.com/cloudfoundry/gorouter/issues/135

As far as I understand, you don't object to this change in principal but this would "involve some substantial changes to how routes are registered".
How can we re-open this issue and get the changes implemented?

This behavior currently causes problems for our customers. Most of our APIs are RESTful. So when a customer sends us, e.g., GET /resources/12345 and gets back 404 Not Found, they assume that this rest resource does not exist and trigger their business logic.  They do not expect that this situation is also possible when our application is temporarily down.

## Steps to Reproduce

Deploy a java application and register a route.
Stop the application. (Any other way of killing the application produces the same result.)
Client sends an http request to an application's endpoint.

## Expected result

Client should receive 503 Service Unavailable.

## Current result

Client receives 404 Not Found.
(To be precise, in a few seconds after the app goes down, the client receives 502 Bad Gateway and 503 Service Unavailable, but later it always receives 404 Not Found).

kecirlotfi:
This PR fixes #250
It depends on the PR cloudfoundry/gorouter#314
A new param config was added to gorouter config file to prevent deleting empty pools when an app is crashed.


jobs/gorouter/spec
```diff
@@ -438,6 +438,13 @@ properties:
       When it set to true, the gorouter will return a 503 when a route is known but there is nothing in the pool available to handle the request. When false, the gorouter will return a 404.
     default: true
 
+  for_backwards_compatibility_only.empty_pool_timeout:
+    description: |
+      This property depends on empty_pool_response_code_503. if empty_pool_response_code_503 is set to true, then gorouter will consider empty_pool_timout parameter.
+      If set to 0, gorouter will remove empty pools after pruning their endpoints. The gorouter will respond with 404 code.
+      If set to different value, gorouter will wait until this timeout is reached then will delete the empty pools. Before pool deletion, gorouter will continue to respond with 503 code.
+    default: 10
```
domdom82: default should be 0 as this is the behavior now (prune pool immediately)

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -387,4 +388,4 @@ if_p('router.html_error_template') do |t|
 end
 
 params.to_yaml[3..-1]
-%>
+%>
```
domdom82: please re-add the newline at the end of the file

jobs/gorouter/spec
```diff
@@ -438,6 +438,13 @@ properties:
       When it set to true, the gorouter will return a 503 when a route is known but there is nothing in the pool available to handle the request. When false, the gorouter will return a 404.
     default: true
 
+  for_backwards_compatibility_only.empty_pool_timeout:
+    description: |
+      This property depends on empty_pool_response_code_503. if empty_pool_response_code_503 is set to true, then gorouter will consider empty_pool_timout parameter.
```
domdom82: if -> If

jobs/gorouter/spec
domdom82: The value is parsed as `time.Duration` in [the PR](https://github.com/cloudfoundry/gorouter/pull/314/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R291) so putting a default of `10` actually means `10ns` (the internal value of time.Duration).

Since we want 0 to be the default, you could either put `0s` as the default or put an example like
```
example: 10s
```
next to the default, so the people will know this is not an integer value but a duration string.

Maybe even add both?

spec/gorouter_templates_spec.rb
```diff
@@ -226,7 +226,8 @@
           'port' => 3745
         },
         'for_backwards_compatibility_only' => {
-          'empty_pool_response_code_503' => true
+          'empty_pool_response_code_503' => true,
+          'empty_pool_timeout' => 10
```
domdom82: here we should also put a dimension in to make it clear we are talking about a time.Duration element.
e.g.
```
'empty_pool_timeout' => '10s'
```

spec/gorouter_templates_spec.rb
```diff
@@ -1018,6 +1019,7 @@
             end
             it 'is set to true' do
               expect(parsed_yaml['empty_pool_response_code_503']).to eq(true)
+              expect(parsed_yaml['empty_pool_timeout']).to eq(10)
```
domdom82:
```suggestion
              expect(parsed_yaml['empty_pool_timeout']).to eq('10s')
```

domdom82 (changes requested):

jobs/gorouter/spec
kecirlotfi: Good point. I've updated that and added some explanation about the type of this parameter.

jobs/gorouter/templates/gorouter.yml.erb
kecirlotfi: Done.

domdom82 (approved):

ameowlia:
🎉  Thank you everyone for your hard work on this!

PR #290, opened 2022-10-06
Migrate router.max_header_bytes to router.max_header_kb

geofffranks:
Replaces painful `router.max_header_bytes` property with slightly less math intensive `router.max_header_kb`.

### DIFF
--- a/jobs/gorouter/templates/gorouter.yml.erb
+++ b/jobs/gorouter/templates/gorouter.yml.erb
@@ -28,11 +28,11 @@ def parse_ip (ip, var_name)
   end
 end
 
-def validate_max_header_bytes (bytes)
-  if bytes < 1 or bytes > 1024 * 1024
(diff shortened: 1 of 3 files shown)


jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -28,11 +28,11 @@ def parse_ip (ip, var_name)
   end
 end
 
-def validate_max_header_bytes (bytes)
-  if bytes < 1 or bytes > 1024 * 1024
-    raise "Invalid router.max_header_bytes value. Must be between 1 and 1,048,576 bytes"
+def validate_max_header_kb (bytes)
```
dsabeti: I think the logic in this function is correct, but I think the `bytes` variable should now be called `kb` or `kilobytes`. When that value is multiplied 1024, _then_ you get the number of bytes.

dsabeti:
@geofffranks I left one comment about a variable name, but otherwise looks good to me!

dsabeti (approved):

PR #293, opened 2022-10-19
remove important ports from ephemeral range

### LINKED ISSUE #206: ip_local_reserved_ports should be set to deconflict with ip_local_port_range
## Issue

If you're unlucky, an unrelated process on a Gorouter can consume a local port that the Gorouter needs to be able to listen on, causing the Gorouter to be stuck in a restart loop.

## Affected Versions

Likely all in the past 5 years.

## Context

https://github.com/cloudfoundry/routing-release/commit/e00e0ca5f8ae3cc9f02abc81642b45a6a38f648d adds the ability to configure `/proc/sys/net/ipv4/ip_local_port_range`, defaulting to `1024 65535` which seems to match what has been there since at least 5 years (!) - I couldn't trace it any further than: https://github.com/cloudfoundry/routing-release/commit/af09610f04e6430c0acdb5206b2aa2ebf226724a#diff-504a89f0b7db13eb0185c6c16ec7ee5c8822573f160db84b7fa56138baf11930R24

Anyway 1024 as a lower bound is much lower than typical defaults which seem to usually start at around 32768 (e.g.: https://tldp.org/LDP/solrhe/Securing-Optimizing-Linux-RH-Edition-v1.3/chap6sec70.html) - however the reasons for doing so seem reasonable enough.

Unless... one doesn't also exclude from that local port range any ports that our VM needs to listen on.

Today we had a Gorouter go bad at system startup time - it wasn't able to bind to the port we had specified it to listen on, because an unrelated process on the same VM had chosen to use that port number as a ephemeral point from the local port range, for a long-lived connection.

This meant that the Gorouter kept (unsuccessfully) trying to restart.

Although a default install currently listens on port 80, which is outside of the local port range (in our case we had overridden that to larger value, e.g. 4080), the default config suggests that other ports are equally likely to be affected, e.g.:

https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/spec#L41-L43

or

https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/spec#L330-L332

## Possible Fix

https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt describes `ip_local_reserved_ports`, pasted below.

If this were set a the same time, and in the same manner as `ip_local_port_range` (including configurable by operators), then this should prevent such a situation from occurring:

```
ip_local_port_range - 2 INTEGERS
	Defines the local port range that is used by TCP and UDP to
	choose the local port. The first number is the first, the
	second the last local port number.
	If possible, it is better these numbers have different parity
	(one even and one odd value).
	Must be greater than or equal to ip_unprivileged_port_start.
	The default values are 32768 and 60999 respectively.

ip_local_reserved_ports - list of comma separated ranges
	Specify the ports which are reserved for known third-party
	applications. These ports will not be used by automatic port
	assignments (e.g. when calling connect() or bind() with port
	number 0). Explicit port allocation behavior is unchanged.

	The format used for both input and output is a comma separated
	list of ranges (e.g. "1,2-4,10-10" for ports 1, 2, 3, 4 and
	10). Writing to the file will clear all previously reserved
	ports and update the current list with the one given in the
	input.

	Note that ip_local_port_range and ip_local_reserved_ports
	settings are independent and both are considered by the kernel
	when determining which ports are available for automatic port
	assignments.

	You can reserve ports which are not in the current
	ip_local_port_range, e.g.:

	$ cat /proc/sys/net/ipv4/ip_local_port_range
	32000	60999
	$ cat /proc/sys/net/ipv4/ip_local_reserved_ports
	8080,9148

	although this is redundant. However such a setting is useful
	if later the port range is changed to a value that will
	include the reserved ports.

	Default: Empty
```

As a workaround, we'll likely use an `osconf` workaround to set that at start-up.

It is satisfying to file this bug - we've observed over the years numerous but infrequent Gorouter failures at startup, but weren't able to pin-point a reason why.  Today we were able to see another process with `lsof` connected to something else using our port, which led us down this path of discovery. Credit to @bg-govau

ameowlia:
Fixes https://github.com/cloudfoundry/routing-release/issues/206 and https://github.com/cloudfoundry/loggregator-agent-release/issues/134

### DIFF
--- a/jobs/gorouter/templates/pre-start.erb
+++ b/jobs/gorouter/templates/pre-start.erb
@@ -6,9 +6,32 @@ source /var/vcap/packages/routing_utils/syslog_utils.sh
 
 tee_output_to_sys_log "${LOG_DIR}" "pre-start" <%= p("router.logging.format.timestamp") %>
 
+<%
+    ports = []
+    ports.append(p("router.port")) # has default. will always exist.
+    ports.append(p("router.status.port")) # has default. will always exist.
+    ports.append(p("router.tls_port")) # has default. will always exist.
+
+    if_p('router.prometheus.port') do |port|
+        ports.append(port)
+    end
+
+    # ipv4 only. This makes me sad. I couldn't find a ruby package that takes an IP
+    # with a port and returns just the port.
+    debug_address = p("router.debug_address").split(":").last.to_i
+    if debug_address > 0
+        ports.append(debug_address)
+    end
+
+    ports.sort!
+%>
+
 <% if p('router.set_kernel_parameters') %>
     # Ephemeral port range
     echo "<%= p("router.ip_local_port_range") %>" > /proc/sys/net/ipv4/ip_local_port_range
+
+    # Reserved Ports, not ephemeral
+    echo <%= ports.join(",") %> > /proc/sys/net/ipv4/ip_local_reserved_ports
 
     # TCP_FIN_TIMEOUT
     # This setting determines the time that must elapse before TCP/IP can release a closed connection and reuse
--- a/jobs/gorouter/templates/pre-start.erb
+++ b/jobs/gorouter/templates/pre-start.erb
@@ -8,6 +8,26 @@ tee_output_to_sys_log "${LOG_DIR}" "pre-start" <%= p("router.logging.format.time
 
 <%
     ports = []
+
+    # the following ports are not available via bosh links and are fragile
+    # or they are available via a link, but they are in an addon, so links don't work
+    bosh_agent_port = 2825
+    bosh_dns_ports = [8853, 53080]
+    forwarder_agent_port = 3458
+    loggregator_agent_port = 3459
+    metrics_agent_port = 3461
+    monit_port = 2822
+    syslog_agent_port = 3460
+
+    ports.append(bosh_agent_port)
+    ports.concat(bosh_dns_ports)
+    ports.append(forwarder_agent_port)
+    ports.append(loggregator_agent_port)
+    ports.append(monit_port)
+    ports.append(syslog_agent_port)
+    ports.append(metrics_agent_port)
+
+    # the following ports are set in this release
     ports.append(p("router.port")) # has default. will always exist.
     ports.append(p("router.status.port")) # has default. will always exist.
     ports.append(p("router.tls_port")) # has default. will always exist.
@@ -31,7 +51,7 @@ tee_output_to_sys_log "${LOG_DIR}" "pre-start" <%= p("router.logging.format.time
     echo "<%= p("router.ip_local_port_range") %>" > /proc/sys/net/ipv4/ip_local_port_range
 
     # Reserved Ports, not ephemeral
-    echo <%= ports.join(",") %> > /proc/sys/net/ipv4/ip_local_reserved_ports
+    echo "<%= ports.join(",") %>" > /proc/sys/net/ipv4/ip_local_reserved_ports
 
     # TCP_FIN_TIMEOUT
     # This setting determines the time that must elapse before TCP/IP can release a closed connection and reuse
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -434,7 +434,7 @@ properties:
       Optimize kernel networking settings for GoRouter. If you deploy GoRouter in a container set this to false.
       This configures the following kernel parameters:
         * Increases the number of allowed IP local port ranges specified by the `router.ip_local_port_range` value.
-        * Reserves ports configured for gorouter via /proc/sys/net/ipv4/ip_local_reserved_ports.
+        * Reserves ports configured for gorouter and other bosh components via /proc/sys/net/ipv4/ip_local_reserved_ports.
         * Reduces the timeout of TCP_FIN_TIMEOUT, so that TCP/IP can release closed connections faster, making more
           resources available for new connections.
         * Disables TCP_TW_RECYCLE to enable fast recycling of TIME_WAIT sockets.
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -434,6 +434,7 @@ properties:
       Optimize kernel networking settings for GoRouter. If you deploy GoRouter in a container set this to false.
       This configures the following kernel parameters:
         * Increases the number of allowed IP local port ranges specified by the `router.ip_local_port_range` value.
+        * Reserves ports configured for gorouter via /proc/sys/net/ipv4/ip_local_reserved_ports.
         * Reduces the timeout of TCP_FIN_TIMEOUT, so that TCP/IP can release closed connections faster, making more
           resources available for new connections.
         * Disables TCP_TW_RECYCLE to enable fast recycling of TIME_WAIT sockets.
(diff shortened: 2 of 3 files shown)


jobs/gorouter/templates/pre-start.erb
```diff
@@ -6,10 +6,33 @@ source /var/vcap/packages/routing_utils/syslog_utils.sh
...
 
+<%
+    ports = []
+    ports.append(p("router.port")) # has default. will always exist.
+    ports.append(p("router.status.port")) # has default. will always exist.
+    ports.append(p("router.tls_port")) # has default. will always exist.
+
+    if_p('router.prometheus.port') do |port|
+        ports.append(port)
+    end
```
geofffranks, line 37: This works for some of the ports we're listening on, but what about other processes that we may or may not be in control of? e.g. a bosh-addon like bosh-dns or an antivirus daemon? 

Should we just block off anything under 10240 and make exceptions for 17001/debug ports?

jobs/gorouter/templates/pre-start.erb
```diff
@@ -6,10 +6,33 @@ source /var/vcap/packages/routing_utils/syslog_utils.sh
...
+
+    ports.sort!
+%>
+
 <% if p('router.set_kernel_parameters') %>
     # Ephemeral port range
     echo "<%= p("router.ip_local_port_range") %>" > /proc/sys/net/ipv4/ip_local_port_range
 
+    # Reserved Ports, not ephemeral
+    echo <%= ports.join(",") %> > /proc/sys/net/ipv4/ip_local_reserved_ports
```
geofffranks: Is this really how that works? All the things i'm reading make it seem like a two integer range, not a comma-separated value.

jobs/gorouter/templates/pre-start.erb
ameowlia: 1. Some fix is better than nothing :D 
1. I would prefer to continue optimizing for the most number of connections, rather than block off everything under 10240.
1.  Here's what I have on a default cf-d
   ```
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 0.0.0.0:8853            0.0.0.0:*               LISTEN      17708/bosh-dns-heal
tcp        0      0 127.0.0.1:2039          0.0.0.0:*               LISTEN      17814/gorouter
tcp        0      0 127.0.0.1:53080         0.0.0.0:*               LISTEN      17736/bosh-dns
tcp        0      0 127.0.0.1:2822          0.0.0.0:*               LISTEN      17759/monit-actual
tcp        0      0 127.0.0.1:2825          0.0.0.0:*               LISTEN      1310/bosh-agent
tcp        0      0 169.254.0.2:53          0.0.0.0:*               LISTEN      17736/bosh-dns
tcp        0      0 169.254.0.2:53          0.0.0.0:*               LISTEN      17736/bosh-dns
tcp        0      0 0.0.0.0:8080            0.0.0.0:*               LISTEN      17814/gorouter
tcp        0      0 127.0.0.1:17002         0.0.0.0:*               LISTEN      17814/gorouter
tcp        0      0 127.0.0.1:15821         0.0.0.0:*               LISTEN      18085/discovery-reg
tcp        0      0 127.0.0.1:14823         0.0.0.0:*               LISTEN      17967/forwarder-age
tcp        0      0 127.0.0.1:14822         0.0.0.0:*               LISTEN      18002/syslog-agent
tcp        0      0 127.0.0.1:14821         0.0.0.0:*               LISTEN      18046/prom-scraper
tcp        0      0 127.0.0.1:14829         0.0.0.0:*               LISTEN      17895/udp-forwarder
tcp        0      0 127.0.0.1:14824         0.0.0.0:*               LISTEN      17931/loggregator-a
tcp        0      0 127.0.0.1:14727         0.0.0.0:*               LISTEN      18121/metrics-agent
tcp        0      0 127.0.0.1:3461          0.0.0.0:*               LISTEN      18121/metrics-agent
tcp        0      0 127.0.0.1:3460          0.0.0.0:*               LISTEN      18002/syslog-agent
tcp        0      0 127.0.0.1:3459          0.0.0.0:*               LISTEN      17931/loggregator-a
tcp        0      0 127.0.0.1:3458          0.0.0.0:*               LISTEN      17967/forwarder-age
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      1021/sshd: /usr/sbi
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      17814/gorouter
tcp        0      0 0.0.0.0:14726           0.0.0.0:*               LISTEN      18121/metrics-agent
tcp        0      0 0.0.0.0:443             0.0.0.0:*               LISTEN      17814/gorouter
udp        0      0 127.0.0.1:323           0.0.0.0:*                           1637/chronyd
udp        0      0 127.0.0.1:3457          0.0.0.0:*                           17895/udp-forwarder
udp        0      0 169.254.0.2:53          0.0.0.0:*                           17736/bosh-dns
udp        0      0 169.254.0.2:53          0.0.0.0:*                           17736/bosh-dns
udp        0      0 10.0.1.9:68             0.0.0.0:*                           504/systemd-network
udp        0      0 0.0.0.0:28746           0.0.0.0:*                           17814/gorouter
   ```
  * I think most of the logging+metrics things choose random ephemeral ports.
  * bosh dns uses mostly udp, not tcp, so those udp ports should be safe.
  * the bosh-dns-health port is a [bosh property here](https://github.com/cloudfoundry/bosh-dns-release/blob/17020f0da21490677df966fc0367a4e845e92f9c/jobs/bosh-dns/spec#L148-L150)
  * we could hardcode that one.
4) What do you think about port `2822` and `2825` for `monit-actual` and `bosh-agent`?
5) If other things have problems, we can play whack-a-mole with them and add them to this list.

jobs/gorouter/templates/pre-start.erb
ameowlia: > ip_local_reserved_ports - list of comma separated ranges
	Specify the ports which are reserved for known third-party
	applications. These ports will not be used by automatic port
	assignments (e.g. when calling connect() or bind() with port
	number 0). Explicit port allocation behavior is unchanged.

>	The format used for both input and output is a comma separated
	list of ranges (e.g. "1,2-4,10-10" for ports 1, 2, 3, 4 and
	10). Writing to the file will clear all previously reserved
	ports and update the current list with the one given in the
	input.
	
[Per the kernel docs](https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt).

jobs/gorouter/templates/pre-start.erb
geofffranks: Is there a way to make monit start the gorouter process last, to have a greater chance of those ports being in use by the correct process when the gorouter goes to grab an next ephemeral port for a backend connection?

We should probably add 8853, 2822, 2825, and possibly 3458-3461 as well since those seem to be consistent.

jobs/gorouter/templates/pre-start.erb
geofffranks: oh i see - different proc setting than the range. 👍 neat!

jobs/gorouter/templates/pre-start.erb
ameowlia: >Is there a way to make monit start the gorouter process last, to have a greater chance of those ports being in use by the correct process when the gorouter goes to grab an next ephemeral port for a backend connection?

I don't think so. I think bosh purposefully doesn't make any promises about ordering.

ameowlia: > 8853, 2822, 2825, and possibly 3458-3461 

I'll try to find bosh links for these and hardcode the rest.

PR #296, opened 2022-10-31
Remove uaa-go-client submodule

geofffranks:
Removes the uaa-go-client submodule so `git submodule update --init --recursive` will work


ameowlia:
docs???

ameowlia:
😝

PR #300, opened 2022-11-14
add gorouter as link; add gorouter backends ca to routing-api

reneighbor:
Release #1 for putting communication between gorouter (client) and routing-api (server) with mtls.

This PR simply prepares routing-api to trust gorouter by adding the gorouter backends CA to routing-api's list of trusted CA's. If CF-D/TAS is not updated to provide the gorouter cert in the manifest, there will be no user-facing change in behavior.

For context, route-registrar doesn't support passing certs for client-to-server validation (mtls), so we were not able to provide a cert to gorouter that routing-api already trusts. Fortunately gorouter already presents a cert chain for validation to backends, so we were able to give routing-api the CA for gorouter's pre-existingcerts. In the future, we may want to update route registrar to support supplying certs as part of route registration.

Tracker story: https://www.pivotaltracker.com/story/show/183523942

### DIFF
--- a/spec/routing_api_templates_spec.rb
+++ b/spec/routing_api_templates_spec.rb
@@ -15,7 +15,7 @@ describe 'routing_api' do
   let(:merged_manifest_properties) do
     {
       'routing_api' => {
-        'mtls_ca' => 'the ca cert',
+        'mtls_ca' => 'a ca cert',
         'mtls_server_key' => 'the server key',
         'mtls_server_cert' => 'the server cert',
         'mtls_client_cert' => 'the client cert',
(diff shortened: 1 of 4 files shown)


spec/routing_api_templates_spec.rb
```diff
@@ -59,6 +58,64 @@
...
+            }
+          )
+        ]
+      end
+      it 'does not render the gorouter backends ca cert' do
+        client_ca = template.render(merged_manifest_properties, consumes: links)
+        expect(client_ca.strip).to eq('a ca cert')
+      end
+    end
+    describe 'when the link is present' do
```
ebroberson: This test looks like a duplicate of the one starting at line 62.  Was that intended?

selzoc: It looks to me that the difference is in the test setup, particularly the `router` stanza of the properties

ebroberson: I see that when comparing ~105 to ~89, but ~68 looks exactly the same.  I think the thing that has me hung is that 78-79 and 115-116 are identical.

selzoc: 🤦 oh totally - I agree, the tests starting on line 62 and 99 appear identical.  Thought you were saying 62 and 83 were the same 🙃

reneighbor: Arg - thank you guys!

ebroberson (approved):
Yay no more dupes!

PR #315, opened 2023-04-11
Remove old lager files from gorouter package spec

dsabeti:


ebroberson (approved):
LGTM

winkingturtle-vmw:
@dsabeti I think routing-release could benefit from having a pre-push git-hooks that checks for this kind of errors (e.g. [diego-release](https://github.com/cloudfoundry/diego-release/blob/develop/git-hooks/pre-push))

PR #331, opened 2023-08-08
Add hop_by_hop_headers_to_filter property

winkingturtle-vmw:
With default values set to:
 - X-Forwarded-For
 - X-Forwarded-Proto
 - B3
 - X-B3
 - X-B3-SpanID
 - X-B3-TraceID
 - X-Request-Start
 - X-Forwarded-Client-Cert

Context: https://github.com/cloudfoundry/gorouter/pull/356

# What is this change about?

Continue to forward relevant headers by removing hop-by-hop headers that could interfere with gorouter.

Context:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection
- https://nathandavison.com/blog/abusing-http-hop-by-hop-request-headers

The following list of keywords are removed from Connection header, so that gorouter can continue to utilize those headers when they are set and will continue to forward them down to the relavant handlers.

# Backwards Compatibility

# How should this be tested?

- Tests are added to the two PRs

# Additional Context

# PR Checklist
* [x] I have [run all the unit tests](https://github.com/cloudfoundry/routing-release#in-a-docker-container) using `scripts/run-unit-tests-in-docker`.

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -284,9 +284,16 @@ properties:
     - name: Accept-Ranges
   router.hop_by_hop_headers_to_filter:
     description: |
-        (optional, array value) List of HTTP Headers that are filtered for Hop-By-Hop Connection header.
-        When clients make requests to the gorouter, these HTTP Headers will be automatically be
-        removed from their request's Connection header when gorouter proxies the request.
+        (optional, array value) List of HTTP Headers that are filtered for
+        Hop-By-Hop Connection header.
+        When clients make requests to the gorouter and the gorouter proxies
+        their request, these HTTP Headers will be automatically be removed from
+        the request's Connection header and sent as normal headers to the
+        target backend. This list *also* specificies the Headers that can be
+        returned by the backend; i.e. if a client attempts to send one of these
+        Headers in their Connection Header to gorouter and it's *NOT* filtered,
+        the header will *NOT* be returned to client properly after proxying the
+        request.
     default:
     - X-Forwarded-For
     - X-Forwarded-Proto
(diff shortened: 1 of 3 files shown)


jobs/gorouter/spec
```diff
@@ -282,6 +282,18 @@ properties:
     example:
     - name: X-Vcap-Request-Id
     - name: Accept-Ranges
+  router.hop_by_hop_headers_to_filter:
+    description: |
+       (optional, array value) List of HTTP Headers that are filtered for Hop-By-Hop Connection header
```
jrussett: I think we should expand this description. I'm not quite sure what this means.

Given your comments on the other PR:
> I am happy to change the name from bannedList to anything folks think is more easily understood, but essentially those header values would be banned from Client side on Connection header and are subject to removal. I know for me personally, headersToKeep would make it sound like a property that would need to have every single header used in gorouter.

From: https://github.com/cloudfoundry/gorouter/pull/356#discussion_r1286054653

> I think between the two options, I would prefer headersToKeep probably because the values in this field are to be removed from Connection header instead of kept.

From: https://github.com/cloudfoundry/gorouter/pull/356#discussion_r1287040572


Maybe we could add a description like:
> (optional, array value) List of HTTP Headers that are filtered for Hop-By-Hop Connection header.
When clients make requests to the gorouter, these HTTP Headers will be automatically be removed from their request's `Connection` header when gorouter proxies the request.

jrussett (changes requested):

jobs/gorouter/spec
winkingturtle-vmw: Updated description

jrussett (approved):
I was able to deploy this and see the behavior work as intended.

PR #334, opened 2023-08-16
Reorganize routing-release test scripts

winkingturtle-vmw:
- Since most tests in this release required Linux, we now only support running tests in Docker and Concourse. This reorganization takes advantage of using the same task for running tests locally in docker and remotely in concourse.
- There is a new requirement for running tests. That is wg-app-platform-runtime-ci that will host helper methods for configure_db and congiure_rsyslog and task for running tests under docker.
- Add missing test-coverage for routing-info

Dependent PR: 
- https://github.com/cloudfoundry/wg-app-platform-runtime-ci/pull/7/files

Related PR:
- https://github.com/cloudfoundry/routing-info/pull/5
- https://github.com/cloudfoundry/cf-tcp-router/pull/15
- https://github.com/cloudfoundry/gorouter/pull/357
- https://github.com/cloudfoundry/routing-api-cli/pull/16
- https://github.com/cloudfoundry/multierror/pull/6
- https://github.com/cloudfoundry/route-registrar/pull/35
- https://github.com/cloudfoundry/routing-api/pull/35

---

# What is this change about?

_In your own words, describe this proposed change and the problem it solves. If GitHub has prepopulated any text above here with info from your commit message, please clean it up and account for it in this section._

- [ ] `[Minor Feature/Improvement]`:  the change introduces a new feature or improvement that doesn't alter core behavior

# Backwards Compatibility

_If this is a breaking change, or modifies currently expected behaviors of core functionality (request routing or request logging), how has the change been mitigated to be backwards compatible?_

_Should this feature be considered experimental for a period of time, and allow operators to opt-in? Should this apply immediately to all routing-release deployments?_

# How should this be tested?

_Are there any non-automated tests that should be performed against this change for validation? Please provide steps, and expected results for before + after the change._

# Additional Context

# PR Checklist
* [ ] I have viewed, signed, and submitted the Contributor License Agreement.
* [ ] I have made this pull request to the `develop` branch.
* [ ] I have [run all the unit tests](https://github.com/cloudfoundry/routing-release#in-a-docker-container) using `scripts/run-unit-tests-in-docker`.
* [ ] I have given thought to the backward-compatibility requirements of this change, and listed any mitigations above.


ebroberson (approved):
LGTM

README.md
```diff
@@ -117,36 +125,24 @@ If you do not have `rspec` or `rubocop` installed locally, run
...
+- `/repo/scripts/docker/test.bash gorouter router`: This will only run `router` sub-package tests for `gorouter` package
 
-* If you'd like to run a specific component's tests in a Docker container,
-  the `run-unit-tests` script also takes a package name as an argument:
+The following helper functions are additionally available by sourcing `./scripts/docker/helpers.bash`:
 
-  ```bash
-  ./scripts/run-unit-tests-in-docker gorouter
-  ```
+- `docker_run_routing_release_tests <mysql-8.0(or mysql),mysql-5.7,postgres>`:  This function combines steps described above sequentially.
```
geofffranks: Can we bring back a dockerized run-template-tests helper function/script?

scripts/docker/helpers.bash
```diff
@@ -0,0 +1,7 @@
+function docker_run_routing_release_tests() {
+  local db="${1?Provide DB flavor. e.g e.g mysql-8.0(or mysql),mysql-5.7,postgres}"
+  local this_file_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+  "$this_file_dir/container.bash" "$db" -d
+  docker exec routing-release-docker-container '/repo/scripts/docker/build-binaries.bash'
+  docker exec routing-release-docker-container '/repo/scripts/docker/test.bash'
```
geofffranks: Is there ever a case where we'd want to run `test.bash` without first running `build-binaries.bash` - could these be merged together?

scripts/docker/test.bash
```diff
@@ -0,0 +1,25 @@
...
+  export DEFAULT_PARAMS="ci/routing-release/default-params/run-bin-test/linux.yml"
+  export ENVS="DB=${DB}"
+  export GOFLAGS="-buildvcs=false"
+  /ci/shared/tasks/run-bin-test/task.bash "${sub_package}"
+}
+
+if [[ -n "${1:-}" ]]; then
+  test "src/code.cloudfoundry.org/${1}" "${2:-}"
+else
+  for component in gorouter cf-tcp-router multierror route-registrar routing-api routing-api-cli routing-info; do
```
geofffranks: Should this use [this list of repos](https://github.com/cloudfoundry/wg-app-platform-runtime-ci/pull/7/files#diff-30a20e3d2dd6d8140c12d8a57c0a08d38eb1e81c8ec7408be9a20f7e4c8343f2R5-R11) instead?

Or maybe instead just look at the contents of `src/code.cloudfoundry.org/` and run `test` on each directory there?

scripts/docker/test.bash
```diff
@@ -0,0 +1,25 @@
...
+
+set -eu
+set -o pipefail
+
+function test() {
+  local package="${1:?Provide a package}"
+  local sub_package="${2:-}"
+
+  export DIR=${package}
+  . <(/ci/shared/helpers/extract-default-params-for-task.bash /ci/shared/tasks/run-bin-test/linux.yml)
```
geofffranks, line 11: Is this call needed since we're passing DEFAULT_PARAMS to the run-bin-test script on line 13?

README.md
winkingturtle-vmw: I think `./scripts/template-tests` is still part of the repo and can be executed inside/outside of the docker container.

scripts/docker/helpers.bash
winkingturtle-vmw: Yeah, not all tests require having the binaries in place. It's best to build-binaries once and have them in docker-container, but if one. wants to run tests for [multi-error](https://github.com/cloudfoundry/multierror/compare/main...with-bin/test.bash) you don't have to build binaries. We did expose the test dependencies so that situation[ like this would not happen](https://github.com/cloudfoundry/routing-acceptance-tests/compare/main...with-bin/test.bash#diff-974b281671c7dc2a0a72f098e7a1547432ebc0216eaf9fe95b46cedad8a6ca6bL24-R31) where routing-acceptance-tests was just relying on a really old version of `rtr` instead of forcing `rtr` to be built in case there was a problem with the code-base.

The idea here is to have a recipe for building-binaries per release. This will make it so that other releases don't have to figure out how to build `rtr`. The recipe would be pulled in from the CI for other release. 

For context, diego's inigo builds `gdn` and `gorouter` and I remember fixing a bug with inigo a while back where it was still building `gdn` with runc instead of containerd.

scripts/docker/test.bash
winkingturtle-vmw: Yes, that's a good catch, we should be pulling in the list from `index.yml` instead of the list here.

scripts/docker/test.bash
winkingturtle-vmw: This will load default values for params that are defined globally e.g. verify_go for VERIFICATIONS. The intention here is that the script that would run inside of the docker container would be exactly identical to what would get executed in concourse. In other words, by fetching the latest `tas-runtime-postgres` image and running the tests in docker we would see go versions mismatch between the release and docker container the same way as it would happen on concourse.

`DEFAULT_PARAMS`  is convention by which a release e.g. routing-release would defines default values for params for which is either not defined or would like to be overridden. This is only needed for run-bin-test or build-binaries since those would be the most common thing a contributor would run.

scripts/docker/test.bash
geofffranks: Started looking into fixing this today, but i'm not sure I like the idea of requiring the CI repo to be present in order to run local tests on routing-release. 

We could have it default to auto-generating in routing-release if the CI repo isn't present to find the index.yml file, but then risk a false sense of security between local testing not matching CI testing.

Thinking about auditing, we would want to catch the following:
- we tested fewer components than we expected
- we tested more components than we expected

These would require two separate lists, and each would need to be updated as a component is added. I'm not sure if this is beneficial since if no action is done after adding a new component, it isn't tested, and nothing is brought to our attention. If it is added to one list but not the other, we get alerted to add it to another list. Most likely it would get added to the list of repos to test first, and then alerted to add to the audit list. 

When using testing uses the same hardcoded list used for auditing, the audit becomes:
- the list of components that were tested matches the codebases in the repo that we wanted to be tested

Like you said in the other PR's comments, it's difficult to programatically decide what components are important, since some are submodules from other releases, or different paths entirely (./src/code.cloudfoundry vs ./src). 

It might be easier to update other releases to ensure everything needing to be tested is in src/code.cloudfoundry.org, and don't include components there if they shouldn't be tested (using a tools package to pull them in as vendored dependencies).

Did you have other methodologies for auditing what's being tested that I'm not coming up with?

geofffranks (approved):

scripts/docker/test.bash
peanball: Why is this path hardcoded to `/ci/...`?

PR #355, opened 2023-09-29
Optional Gorouter mTLS Client Certificate Metadata Verification

peanball:
# What is this change about?

We want to add optional mTLS client certificate metadata verificiation in addition to basic validity checks.

Given a particular signing CA (identified by its subject) we want to limit the allowed client certificate subjects.

While the underlying issue is better solved with changes to the PKI and trust relationship between CAs, the PKI or its use by other sub-entities is not always under the control of the operator.

This PR pulls in the Gorouter change that implements the feature and provides the configuration spec and template to configure the optional property.`

# Backwards Compatibility

This is a non-breaking change that is completely optional. When the property is not set, no callback is registered at all.

# How should this be tested?

See Gorouter PR. There are also automated tests as part of the Gorouter PR.

# Additional Context

Gorouter PR: https://github.com/cloudfoundry/gorouter/pull/367

# PR Checklist
* [x] I have [run all the unit tests](https://github.com/cloudfoundry/routing-release#in-a-docker-container) using ~~`scripts/run-unit-tests-in-docker`~~ `scripts/test-in-docker-locally.bash`.

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -518,6 +518,21 @@ properties:
     description: "The number of file descriptors a router can have open at one time"
     default: 100000
 
+  router.verify_client_certificate_metadata:
+    description: |
+      Additional client certificate verification which limits the allowed client certificate for given to a signing CA (identified by its subject) to the certificates with subjects provided in the list of valid subjects. Each list entry contains a ca_subject with a coresponding list of valid subjects.
+      - ca_subject:
+          - common_name: ""
+            organization: []
+            locality: []
+            country: []
+        valid_subjects:
+          - common_name: ""
+            organizational_units: []
+            organizations: []
+            locality: []
+            country: []
+    default: []
   healthchecker.failure_counter_file:
     description: "File used by the healthchecker to monitor consecutive failures."
     default: /var/vcap/data/gorouter/counters/consecutive_healthchecker_failures.count
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -525,8 +525,8 @@ properties:
   router.verify_client_certificate_metadata:
     description: |
       Additional client certificate verification, after the certificate was validated using the regular mTLS mechanism and is issued using one of the CAs in `client_ca_certs`.
-      The additional verification limits the allowed client certificates for a given signing CA (identified by its subject) to certificates with subjects provided in the list of valid subjects. Within the certificate chain there may be more than one CA certificates (e.g. intermediate CA certificates). The `issuer_in_chain` must match one of the CA certificates in the chain.
-      Each list entry contains a issuer_in_chain with a corresponding list of valid subjects. Each issuer_in_chain must match one of the certificates in `client_ca_certs`. When an issuer_in_chain is defined that does not match, this raises an error during templating time and at startup in gorouter.
+      The additional verification limits the allowed client certificates for a given signing CA (identified by its distinguished name) to certificates with subjects provided in the list of valid subjects. Within the certificate chain there may be more than one CA certificates (e.g. intermediate CA certificates). The `issuer_in_chain` must match one of the CA certificates in the chain.
+      Each list entry contains an issuer_in_chain with a corresponding list of valid subjects. Each issuer_in_chain must match one of the certificates in `client_ca_certs`. When an issuer_in_chain is defined that does not match, this raises an error during templating time and at startup in gorouter.
       - issuer_in_chain:
           common_name: ""
           serial_number: ""
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -524,18 +524,29 @@ properties:
     default: false
   router.verify_client_certificate_metadata:
     description: |
-      Additional client certificate verification which limits the allowed client certificate for given to a signing CA (identified by its subject) to the certificates with subjects provided in the list of valid subjects. Each list entry contains a ca_subject with a coresponding list of valid subjects.
+      Additional client certificate verification, after the certificate was validated using the regular mTLS mechanism and is issued using one of the CAs in `client_ca_certs`.
+      The additional verification limits the allowed client certificates for a given signing CA (identified by its subject) to certificates with subjects provided in the list of valid subjects. Within the certificate chain there may be more than one CA certificates (e.g. intermediate CA certificates). The `ca_subject` must match one of the CA certificates in the chain.
+      Each list entry contains a ca_subject with a corresponding list of valid subjects. Each ca_subject must match one of the certificates in `client_ca_certs`. When a ca_subject is defined that does not match, this raises an error during templating time and at startup in gorouter.
       - ca_subject:
-          - common_name: ""
-            organization: []
-            locality: []
-            country: []
+          common_name: ""
+          serial_number: ""
+          country: []
+          organisation: []
+          organisation_unit: []
+          locality: []
+          province: []
+          street_address: []
+          postal_code: []
         valid_subjects:
           - common_name: ""
-            organizational_units: []
-            organizations: []
-            locality: []
+            serial_number: ""
             country: []
+            organisation: []
+            organisation_unit: []
+            locality: []
+            province: []
+            street_address: []
+            postal_code: []
     default: []
   healthchecker.failure_counter_file:
     description: "File used by the healthchecker to monitor consecutive failures."
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -518,6 +518,10 @@ properties:
     description: "The number of file descriptors a router can have open at one time"
     default: 100000
 
+  router.enable_verify_client_certificate_metadata:
+    description: |
+        Enable additional client certificate verification via verify_client_certificate_metadata (see below).
+    default: false
   router.verify_client_certificate_metadata:
     description: |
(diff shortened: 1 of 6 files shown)


jobs/gorouter/spec
```diff
@@ -518,6 +518,25 @@ properties:
     description: "The number of file descriptors a router can have open at one time"
     default: 100000
 
+  router.enable_verify_client_certificate_metadata:
+    description: |
+        Enable additional client certificate verification via verify_client_certificate_metadata (see below).
+    default: false
+  router.verify_client_certificate_metadata:
+    description: |
+      Additional client certificate verification which limits the allowed client certificate for given to a signing CA (identified by its subject) to the certificates with subjects provided in the list of valid subjects. Each list entry contains a ca_subject with a coresponding list of valid subjects.
```
domdom82: `certificate for given ^ to a signing CA`
                                   
there seems to be a word missing

jobs/gorouter/spec
```diff
@@ -518,6 +518,25 @@ properties:
...
+      - ca_subject:
+          - common_name: ""
```
domdom82, line 541: so a ca_subject can have multiple subjects? are we sure this should be a list?

jobs/gorouter/spec
```diff
@@ -518,6 +518,25 @@ properties:
...
+            organization: []
```
domdom82: what about the rest of possible names?
OrganizationalUnit
Province
StreetAddress
PostalCode
SerialNumber

Practically, things like PostalCode or StreetAddress we'll probably never see in real life, however `SerialNumber` is a thing that I've seen change between to otherwise equal CA names.

jobs/gorouter/spec
```diff
@@ -518,6 +518,25 @@ properties:
...
+            locality: []
+            country: []
+        valid_subjects:
+          - common_name: ""
```
domdom82: same here. possible other names:

Province
StreetAddress
PostalCode
SerialNumber

jobs/gorouter/spec
```diff
@@ -518,6 +518,25 @@ properties:
     description: "The number of file descriptors a router can have open at one time"
     default: 100000
 
+  router.enable_verify_client_certificate_metadata:
+    description: |
+        Enable additional client certificate verification via verify_client_certificate_metadata (see below).
+    default: false
+  router.verify_client_certificate_metadata:
```
domdom82, line 525: Basic question: Does the issuer or subject DN have to match exactly?
So, when the incoming request has this structure:
```
issuer=C = US, ST = New Jersey, L = Jersey City, O = The USERTRUST Network, CN = USERTrust RSA Certification Authority
subject=C = DE, ST =BW, O = SAP, CN = CloudFoundry
``` 
and my configured values are:
```
- ca_subject:
       - organization: ["The USERTRUST Network"]
```

will it match or not? 
if it has to match exactly all fields, this could become dangerous if only slight variations of the CA certificate occur.

domdom82 (changes requested):

jobs/gorouter/spec
peanball: it should not be a list, thanks!

jobs/gorouter/spec
peanball: At the moment the complete DN has to match exactly, i.e. partial matches will _not_ be accepted.

Could this be an issue with the SerialNumber field?

jobs/gorouter/spec
domdom82: Another thing that's a bit confusing here:
- We provide a list of `ca_subject` entries, coupled with a list of `valid_subjects`

Usually when speaking about x509 certificates, there is an "issuer" (the CA) and a "subject" (the thing that is signed by the certificate). What may be confusing here is that the "issuer" may as well be a "subject" one step above in the chain where the previous certificate was signed by the preceding CA. However, I would like to make it clear that for the trust check we apply here, we will only ever look at leaf certificates, which will only have one "issuer" DN and one "subject" DN.

So to me it would make more sense to speak of a list of `issuer` entries coupled with a list of "subject" entries.

peanball: The whole chain is inspected, where the certificate's issuer is not necessarily the root CA certificate defined in the `ca_subject`.

Example:
- root CA, issuer "CN=root CA", subject "CN=root CA"
- ICA, issuer "CN=root CA", subject "CN=Intermediate CA"
- Client Cert, issuer "CN=intermediate CA", subject "CN=Client, Inc."

You'd define:
```yaml
- ca_subject:
    common_name: root CA
  valid_subjects:
    - common_name: Client, Inc.
```

You _could_ also define a ca_subject with the intermediate CA:
```yaml
- ca_subject:
    common_name: intermediate CA
  valid_subjects:
    - common_name: Client, Inc.
```

Ulitmately jus the DN of one of the CA certificates (root or intermediate) in the certificate chain must match the given subject. It's not necessarily the client certificate's issuer.

jobs/gorouter/spec
peanball: I'll add those to the example. They're all supported in the code.

jobs/gorouter/spec
peanball: Same as above.

jobs/gorouter/spec
domdom82: my point remains. to me it would make equal sense to use this notation:
```
- issuer:
    common_name: root CA
  valid_subjects:
    - common_name: Client, Inc.
```
this means "if anywhere in the chain we have `root CA` as the issuer, then we will accept `Client, Inc.` as valid subjects of a **leaf** certificate". (we might even rename `valid_subjects` to `valid_leaf_subjects` to make it clearer...

the same goes for the intermediate:
```
- issuer:
    common_name: intermediate CA
  valid_subjects:
    - common_name: Client, Inc.
```
here it becomes even clearer. we mean to say: "if `intermediate CA` is the issuer, we will accept `Client, Inc.` as valid subjects for any **leaf** certificate"

peanball: ok, fair. `ca_subject` exposes an implementation detail, i.e. that we check the CA certificates' subjects in the chain.

Even clearer could be the following?
```yaml
- issuer_in_chain:
    common_name: root CA
  valid_cert_subjects:
    - common_name: Client, Inc.
```

WDYT?

domdom82: Yes! I really like it much better 😁

jobs/gorouter/spec
domdom82: As agreed we will go with a full match for the time being.

jobs/gorouter/spec
```diff
@@ -518,6 +518,36 @@ properties:
...
+    description: |
+      Additional client certificate verification, after the certificate was validated using the regular mTLS mechanism and is issued using one of the CAs in `client_ca_certs`.
+      The additional verification limits the allowed client certificates for a given signing CA (identified by its subject) to certificates with subjects provided in the list of valid subjects. Within the certificate chain there may be more than one CA certificates (e.g. intermediate CA certificates). The `issuer_in_chain` must match one of the CA certificates in the chain.
```
domdom82:
```suggestion
      The additional verification limits the allowed client certificates for a given signing CA (identified by its distinguished name) to certificates with subjects provided in the list of valid subjects. Within the certificate chain there may be more than one CA certificates (e.g. intermediate CA certificates). The `issuer_in_chain` must match one of the CA certificates in the chain.
```

domdom82: one less "subject"...😜

jobs/gorouter/spec
```diff
@@ -518,6 +518,36 @@ properties:
...
+      Each list entry contains a issuer_in_chain with a corresponding list of valid subjects. Each issuer_in_chain must match one of the certificates in `client_ca_certs`. When an issuer_in_chain is defined that does not match, this raises an error during templating time and at startup in gorouter.
```
domdom82:
```suggestion
      Each list entry contains an issuer_in_chain with a corresponding list of valid subjects. Each issuer_in_chain must match one of the certificates in `client_ca_certs`. When an issuer_in_chain is defined that does not match, this raises an error during templating time and at startup in gorouter.
```

spec/gorouter_templates_spec.rb
```diff
@@ -771,6 +771,89 @@
...
+                deployment_manifest_fragment['router']['verify_client_certificate_metadata'] = [
+                  { "issuer_in_chain" => { "common_name" => "test-with-san.com", "country" => ["US"] },
+                    "valid_cert_subjects" => [
+                      {"issuer_in_chain" => { "common_name" => "test.com client cert1" }},
+                      {"issuer_in_chain" => { "common_name" => "test.com client cert2", "locality" => ["US"] }}
+                    ]
+                  }
+                ]
+              end
+              it 'fails and explains the validpopulates the properties after a successful check' do
```
domdom82: this sentence seems odd 🧐

domdom82 (dismissed):
only some minor gripes, approving.

spec/gorouter_templates_spec.rb
peanball:
```suggestion
              it 'fails and explains the valid cert subjects in the message' do
```

domdom82 (approved):

ameowlia (approved):

PR #372, opened 2023-12-15
Add TLS health check for tcp-router

geofffranks:
---

Adds a TLS health check listener for tcp-router, and enables operators to disable non-tls health checks for tcp-router.

- [x] `[Breaking Change]`: the change removes a feature or introduces a behavior change to core functionality (request routing, request logging)
- [ ] `[Minor Feature/Improvement]`:  the change introduces a new feature or improvement that doesn't alter core behavior

Technically a breaking change/not backwards compatible from the routing-release perspective, as new properties have been added without defaults. However they will be provided via variables in cf-deployment.

# Backwards Compatibility

cf-deployment will be updated to provide the new properties.

# How should this be tested?

# Additional Context

https://www.pivotaltracker.com/story/show/186497091
https://www.pivotaltracker.com/story/show/186471705

# PR Checklist
* [x] I have [run all the unit tests](https://github.com/cloudfoundry/routing-release#running-unit-and-integration-tests).

### DIFF
--- a/jobs/tcp_router/spec
+++ b/jobs/tcp_router/spec
@@ -67,6 +67,10 @@ properties:
   tcp_router.request_timeout_in_seconds:
     description: "Server and client timeouts in seconds"
     default: 300
+
+  tcp_router.enable_nontls_health_checks:
+    description: "Toggles on/off whether or not to listen for load balancer health check requests on the non-tls `tcp_router.health_check_port` port"
+    default: true
   tcp_router.health_check_port:
     description: "Load balancer in front of TCP Routers should be configured to check the health of TCP Router instances by establishing a TCP connection on this port"
     default: 80
--- a/jobs/tcp_router/spec
+++ b/jobs/tcp_router/spec
@@ -10,6 +10,7 @@ templates:
   routing_api_client_certificate.crt.erb: config/certs/routing-api/client.crt
   routing_api_client_private.key.erb: config/keys/routing-api/client.key
   routing_api_ca_certificate.crt.erb: config/certs/routing-api/ca_cert.crt
+  tcp_router_health_check_certificate.pem.erb: config/certs/health.pem
   haproxy.conf.erb: config/haproxy.conf
   haproxy.conf.template.erb: config/haproxy.conf.template
   bpm.yml.erb: config/bpm.yml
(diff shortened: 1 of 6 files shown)


jobs/tcp_router/spec
```diff
@@ -66,9 +67,20 @@ properties:
   tcp_router.request_timeout_in_seconds:
     description: "Server and client timeouts in seconds"
     default: 300
+
+  tcp_router.enable_nontls_health_checks:
+    description: "Toggles on/off whether or not to listen for load balancer health check requests on the non-tls `tcp_router.health_check_port` port"
+    default: true
```
maxmoehl, line 73: Couldn't we change the logic a bit to make this a non-breaking change? We could name it more generic (e.g. just `health_checks`) and turn it into an enum with possible values being: `off`, `tls`, `tcp`, `tcp+tls` (or `all`) and defaulting to `tcp` to make the change compatible with previous versions.

maxmoehl (commented):
Is there a specific reason to require TLS instead of making it optional? See also the inline comment.

I can't access the linked stories, sorry if it's already clarified there.

ebroberson (dismissed):
LGTM

jobs/tcp_router/spec
geofffranks: We wanted to make this behave similarly to gorouter's health checks:

- always-on TLS listener
- tcp/http listener on  by default, but able to be configured off for environments wanting only TLS health checks (mainly for making security scanners less noisy)

Realistically we could do this in a non-breaking way, but adding the cert configuration to the base cf-deployment yml is a lot easier than adding checks to ensure that at least one of tls or tcp is enabled in both gorotuer + tcp router.

ameowlia: @maxmoehl - When this feature is released in cf-d we will make sure to also create and provide a cert at the same time, so that this will **NOT** be a breaking change for cf-d users.

If we take these precautions to not break users, and we model our work after how this is done in gorouter, do you have other concerns about breaking changes?

ameowlia (approved):

jobs/tcp_router/spec
maxmoehl: Hm, personally I don't see why you would make a health-check run with TLS. To me this is just a waste of resources given that in bigger environments _a lot_ of health checks are happening and you don't want keep-alive on health checks => many wasted CPU cycles for (IMHO) pointless TLS handshakes.

So at least in our environments I wouldn't want to enable TLS health checks but with this change we still have to provide the certificates for it (well cf-d provides them but they need to be generated, rotated, ...).

Bottom line: I'd prefer if we have equal support for both cases across all components. I've missed [the PR to gorouter](https://github.com/cloudfoundry/routing-release/pull/375) but would've raised a similar comment. (But I must also admit that I currently don't have the time to look at this in-depth right now)

maxmoehl: Okay I've realised that while the certs are added the non-tls health-check is still enabled by default. Since the PR is merged anyways I'll go ahead and resolve this.

PR #382, opened 2024-01-25
feat: Add route_services_internal_server_port property.

### LINKED ISSUE #342: Gorouter should use a dedicated port for local route service requests
## Is this a security vulnerability?
No

## Issue
While debugging into a different problem on Gorouter, we stumbled over seemingly random messages like these:
```
read tcp 127.0.0.1:26057->127.0.0.1:19244: read: connection reset by peer
```
as we were digging further we concluded that gorouter would talk to itself on a random port that it uses for re-injecting route service requests into itself, i.e. accept client request -> look up target route -> realize there is a route service for that route -> send a new request to self where the target URL replaced with route service URL

We find it to be very confusing in the logs that gorouter uses a random local port and would like to suggest to use a well-defined port instead. 

## Affected Versions
All

## Context

The random port is selected in the [RouteServicesServer](https://github.com/cloudfoundry/gorouter/blob/main/router/route_service_server.go#L53) where it is then stored in the [RouteServiceRoundTripper](https://github.com/cloudfoundry/gorouter/blob/main/router/route_service_server.go#L98) for future round trips.

## Steps to Reproduce

1. Push an app
2. Push a custom route service
3. Register the route service for the app's route
4. Curl the app's URL
5. Log on to gorouter
6. Run
```
netstat -pant | grep 127
```

## Expected result

There should be a connection from 127.0.0.1:< random port > to 127.0.0.1:< well-defined port >

## Current result

There is a connection from 127.0.0.1:< random port > to 127.0.0.1:< random other port >

## Possible Fix

- We should agree on a port (e.g. 7070)
- The port should not clash with random local ports (i.e. outside gorouter's local port range, similar to 8080)
- BTW the [default port range](https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/spec#L485) is actually 1024-65535 which is potentially dangerous as it allows port 8080 to be used which clashes with the health check port. We may consider using `net.ipv4.ip_local_reserved_ports` to protect those ports from being used by Linux.
- The port should be present as a constant in [gorouter code](https://github.com/cloudfoundry/gorouter/blob/main/router/route_service_server.go) and documented wherever necessary

## Additional Context

If there are strong reasons why the port must be random, please let me know in the discussion of this issue.

domdom82:
# What is this change about?

Fixes #342 

- Removes randomly selected port for internal route service server
- Sets a dedicated port (default: 7070)

# Backwards Compatibility

_If this is a breaking change, or modifies currently expected behaviors of core functionality (request routing or request logging), how has the change been mitigated to be backwards compatible?_
Yes. The port change should be transparent as it is only used internally and only opened on localhost.

_Should this feature be considered experimental for a period of time, and allow operators to opt-in? Should this apply immediately to all routing-release deployments?_
No.

# How should this be tested?

_Are there any non-automated tests that should be performed against this change for validation? Please provide steps, and expected results for before + after the change._
- Spec tests have been provided
- Unit and integration tests have been adjusted in companion PR

# Additional Context

- Companion [Gorouter PR](https://github.com/cloudfoundry/gorouter/pull/389)

# PR Checklist
* [x] I have [run all the unit tests](https://github.com/cloudfoundry/routing-release#running-unit-and-integration-tests).

### DIFF
--- a/spec/gorouter_templates_spec.rb
+++ b/spec/gorouter_templates_spec.rb
@@ -1509,7 +1509,7 @@ describe 'gorouter' do
       { 'router' => {
         'port' => 81,
         'status' => { 'port' => 8081, 'tls' => {'port' => 8443}, },
-        'prometheus' => { 'port' => 7070 },
+        'prometheus' => { 'port' => 7777 },
         'tls_port' => 442,
         'debug_address' => '127.0.0.1:17003'
       } }
@@ -1518,7 +1518,7 @@ describe 'gorouter' do
     context 'ip_local_reserved_ports' do
       it 'contains reserved ports in order' do
         rendered_template = template.render(properties)
-        ports = '81,442,2822,2825,3457,3458,3459,3460,3461,7070,8081,8082,8443,8853,9100,14726,14727,14821,14822,14823,14824,14829,14830,14922,15821,17003,53035,53080'
+        ports = '81,442,2822,2825,3457,3458,3459,3460,3461,7070,7777,8081,8082,8443,8853,9100,14726,14727,14821,14822,14823,14824,14829,14830,14922,15821,17003,53035,53080'
         expect(rendered_template).to include("\"#{ports}\" > /proc/sys/net/ipv4/ip_local_reserved_ports")
       end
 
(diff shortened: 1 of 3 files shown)


spec/gorouter_templates_spec.rb
```diff
@@ -1509,7 +1509,7 @@
       { 'router' => {
         'port' => 81,
         'status' => { 'port' => 8081, 'tls' => {'port' => 8443}, },
-        'prometheus' => { 'port' => 7070 },
+        'prometheus' => { 'port' => 7777 },
```
maxmoehl, line 1512: Let's hope this was just used in tests and no-one has this configured on a prod landscape. We should probably add a note to the release notes.

maxmoehl (approved):

spec/gorouter_templates_spec.rb
domdom82: true. it's a freely configurable value so we should be good. I'll add it to the release notes of the next release

PR #407, opened 2024-05-08
feat: Add per-route options to route_registrar spec

b1tamara:

Summary
- Add route-registrar spec property for custom per-route option load balancing algorithm. It is possible to choose between `round-robin` and `least-connection`.
- Implementation of custom options in route-registrar: [route-registrar#40](https://github.com/cloudfoundry/route-registrar/pull/40)
- Change according to [RFC](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0027-generic-per-route-features.md)

Backward Compatibility
---------------
Breaking Change? **No**


domdom82 (approved):
LGTM

PR #410, opened 2024-05-14
Use forked negroni/v3 to work around HTTP Status Code issue  with 100 continues

### LINKED ISSUE #409: Implementing header "Expect: 100-continue" breaks with routing-release >= 0.293.0
### Behavior
If a CF app is implementing the `Expect: 100-continue` header to first read all headers of a request before reading the body, the respective scenario is breaking with routing-release higher than 0.293.0. Previously, the CF app would respond with `201 Created`, whereas this request is responded with `200 ok` [by gorouter](https://github.com/golang/go/blob/master/src/net/http/server.go#L1639) in the new version. 

### Steps to reproduce
* push this simple app that reads the headers (and, as an example, checks the size of content-length before actually reading the request body): https://github.com/plowin/expect-100
* send a request to the /expect handler comparing 0.292.0 and >=0.293.0 and compare the responses, 201 and 200
* `curl` did some magic and the header went missing, but you can see the behavior when e.g. using "bruno":
0.292.0: 
0.296.0: 

### References
* This seems to be an undesired side-effect of https://github.com/cloudfoundry/gorouter/pull/395.

### Desired behavior

* The default behavior is restored and clients can continue using 100-continue as before
* A test is implemented in routing-release to test the behavior and avoid such regressions in the future

### Affected Version

0.293.0

geofffranks:

Summary
---------------
Needed until https://github.com/urfave/negroni/pull/277 is merged, to resolve https://github.com/cloudfoundry/routing-release/issues/409.

Backward Compatibility
---------------
Breaking Change? **Yes/No**
No, fixes a regression.

### DIFF
--- a/src/code.cloudfoundry.org/go.mod
+++ b/src/code.cloudfoundry.org/go.mod
@@ -10,6 +10,9 @@ replace github.com/codegangsta/cli => github.com/codegangsta/cli v1.6.0
 
 replace github.com/cactus/go-statsd-client => github.com/cactus/go-statsd-client v2.0.2-0.20150911070441-6fa055a7b594+incompatible
 
+// Use forked negroni to work around https://github.com/cloudfoundry/routing-release/issues/409 until https://github.com/urfave/negroni/pull/277 is merged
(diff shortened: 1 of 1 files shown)


jrussett (approved):
Talked to Geoff and he said he'd make a story to track removing this replace directive.

PR #422, opened 2024-06-27
Add TLS/mTLS support to TCP Router

Birdrock:

Summary
---------------
Adds TLS/mTLS support to TCP Router.

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/jobs/tcp_router/templates/tcp_router.yml.erb
+++ b/jobs/tcp_router/templates/tcp_router.yml.erb
@@ -59,9 +59,9 @@ haproxy_pid_file: "/var/vcap/data/tcp_router/config/haproxy.pid"
 isolation_segments: <%= p("tcp_router.isolation_segments") %>
 reserved_system_component_ports: <%=  reserved_system_component_ports %>
 <%
-    ca_cert =     p('tcp_router.backend_tls.ca_cert', '')
-    client_cert = p('tcp_router.backend_tls.client_cert', '')
-    client_key =  p('tcp_router.backend_tls.client_key', '')
+    ca_cert =     p('tcp_router.backend_tls.ca_cert', '').strip
+    client_cert = p('tcp_router.backend_tls.client_cert', '').strip
+    client_key =  p('tcp_router.backend_tls.client_key', '').strip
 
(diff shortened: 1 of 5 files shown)


geofffranks (commented):
BOSH link support was not added for pulling in the CA, because gorouter does not use links to pull in the CA for its backend tls info, and we were mimicking its behavior here.

Additionally rep does not provide a link with its executor's instance identity CA.

jobs/tcp_router/templates/tcp_router.yml.erb
```diff
@@ -58,3 +58,29 @@ routing_api:
 haproxy_pid_file: "/var/vcap/data/tcp_router/config/haproxy.pid"
 isolation_segments: <%= p("tcp_router.isolation_segments") %>
 reserved_system_component_ports: <%=  reserved_system_component_ports %>
+<%
```
ameowlia, line 61: Sometimes people accidentally put in spaces or new lines for certs when they don't actually want anything set. Esp when using GUIs to set these properties. 

❓ Can we add a space/newline trim to this logic? And if there were only spaces then it should treat it as if nothing was provided?

ameowlia (changes requested):

jobs/tcp_router/templates/tcp_router.yml.erb
Birdrock: We definitely can. The block is already trimming trailing white space, so we can trim leading, as well.

One thought is that it might be using a lack of trimming to provide a new line.

Birdrock: This looks like an execution block. Is the request to trim `ca_cert`, `client_cert`, and `client_key` individually? Since the values are there to check for various permutations, but aren't part of the output, do we still want to strip leading white space?

geofffranks: Correct, i think we just want to call `.strip` when setting those three variables.

geofffranks:
Closing as this code was included in #425 with a couple extra commits that addressed PR feedback here + @Birdrock is temporarily on another project and we can't push to his branch.

PR #431, opened 2024-07-31
add additional unsafe params

pnikonowicz:

Summary
---------------
this allows for more unsafe params to be set.

this is to allow postgres to invoke and access
data persistent disks which are necessary in order to query the database to determine if a node is
primary during an HA configuration.

allows for privlaged and additional volumes
to be mounted

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/jobs/route_registrar/spec
+++ b/jobs/route_registrar/spec
@@ -132,9 +132,15 @@ properties:
         name (required, string): Human-readable reference for the healthcheck
         script_path (required, string): Path to script that will be run periodically to determine
           service health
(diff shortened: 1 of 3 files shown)


ameowlia:
Hi @pnikonowicz , Thanks for submitting this. 

❓ Can you provide steps so I can deploy and test this behavior?
❓ Can you provide more information about why this is needed now, but was never needed before?

ameowlia (approved):
✅ read through code
✅ ran tests locally
✅ deployed code
✅ since bpm.yml template tests didn't exist before, I manually checked with a cf-d deployment that the configs didn't change. 

Holding off on merging until the pipeline is green again.

PR #433, opened 2024-08-19
Add dynamic_config_globs property to route_registrar

mariash:

Summary
---------------
Add new property `dynamic_config_globs` to route-registrar job that specifies list of globs route-registrar will be scanning periodically to register routes dynamically.
Also, simplify bpm template, remove lambda from unrestricted volumes.

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/jobs/route_registrar/templates/bpm.yml.erb
+++ b/jobs/route_registrar/templates/bpm.yml.erb
@@ -22,12 +22,14 @@ paths = []
 unrestricted_volumes = []
 routes = p('route_registrar.routes')
 unsafe_privileged = false
+
 routes.each do |route|
   if route['health_check']
     # valid path is /var/vcap/jobs/JOB
     matched = /(^\/var\/vcap\/jobs\/[^\/]*)\/.*/.match(route['health_check']['script_path'])
     if matched
-      paths << matched[1]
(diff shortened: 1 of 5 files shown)


tcdowney:
cc @steven-zou

jobs/route_registrar/templates/bpm.yml.erb
```diff
@@ -22,15 +22,17 @@ paths = []
...
     # valid path is /var/vcap/jobs/JOB
     matched = /(^\/var\/vcap\/jobs\/[^\/]*)\/.*/.match(route['health_check']['script_path'])
     if matched
-      paths << matched[1]
+      paths.append({ "path" => matched[1], "allow_executions" => true })
+      paths.append({ "path" => matched[1].sub('jobs', 'data') })
     end
 
-    if route['health_check']['privileged']
+    if unsafe_privileged == false && route['health_check']['privileged']
```
geofffranks: Why does this need to check unsafe_privileged now? line 24 sets it explicitly to false so boolean would always evaluate to true at this point in the template

mariash: @geofffranks yeah, I misread it. I thought they loop over routes and overwrite with the latest value. This first check can go away.

geofffranks (approved):

PR #437, opened 2024-09-26
fix: bump minimum go version to 1.23 for partitioned cookies

maxmoehl:

Summary
---------------

Bump minimum go version to 1.23 since support for partitioned cookies was only introduced with go1.23.

Backward Compatibility
---------------
Breaking Change? **No**


maxmoehl:
See also: cloudfoundry/gorouter#441

MarcPaquette (approved):

PR #443, opened 2024-11-12
feat: add support for additional request limits

### LINKED ISSUE #309: [gorouter] Add property to restrict header count and size
We deploy gorouter behind HAProxy which limits the amount of headers and the size of each header for requests and responses. For requests towards CF this isn't much of an issue since the request is rejected with `400: Bad Request` and it never reaches the application. However, response headers are processed by the gorouter and the request shows up as `200: OK` but is later rejected by HAProxy and returned as a `502: Bad Gateway` to the client.

To avoid this behaviour which is confusing for users, we would like to introduce two properties:
- `max_headers: int` to limit the amount of req/res headers
- `max_header_size: int` to limit the number of bytes ~~a single header is allowed to occupy~~ the request/response line and headers (without body) are allowed to occupy

This allows us to reject responses in gorouter and show the error message in the app logs which are visible to the user.

We can implement this feature but would like to first discuss if there are any concerns/questions. Leave a :+1: if you like this / don't see any issues with it.

maxmoehl:
Resolves: https://github.com/cloudfoundry/routing-release/issues/309

Summary
---------------
See also: https://github.com/cloudfoundry/gorouter/pull/453

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -267,11 +267,40 @@ properties:
       default: false
   router.max_header_kb:
     description: |
+        Deprecated, use router.max_request_header_kb instead which is equivalent to this option.
+    default: 1024 # 1Mb
+  router.max_request_header_kb:
+    description: |
         This value controls the maximum number of bytes (in KB) the gorouter will read
         parsing the request header's keys and values, including the request
         line. It does not limit the size of the request body. Requests with
         larger headers will result in a 431 status code. Must be between 1 and 1024kb.
-    default: 1024 # 1Mb
+        Note: This value takes precedence over router.max_header_kb but has no default to not break
+        existing setups. If you previously configured router.max_header_kb it is recommended to
+        switch to this property instead.
+    example: 1024 # 1Mb
+  router.max_response_header_kb:
+    description: |
+        This value controls the maximum number of bytes (in KB) the gorouter will read
+        parsing the response header's keys and values, including the request
+        line. It does not limit the size of the response body. Responses with
+        larger headers will result in a 502 status code. A limit of zero or less will
+        result in the default GoLang limit being used.
+    default: 0
+  router.max_request_headers:
+    description: |
+        This value controls the maximum number of headers gorouter will accept in a
+        single request. Only the header keys are counted and values separated by commas
+        are not considered additional headers. Setting this to zero or less disables the
+        limit. Must be at least 30 for CF to function properly (enforced).
+    default: 0
+  router.max_response_headers:
+    description: |
+        This value controls the maximum number of headers gorouter will accept in a
+        single response. Only the header keys are counted and values separated by commas
+        are not considered additional headers. Setting this to zero or less disables the
+        limit. Must be at least 30 for CF to function properly (enforced).
+    default: 0
   router.extra_headers_to_log:
     description: "An array of headers that access log events will be annotated with. This only applies to headers on requests."
     default: []
--- a/jobs/gorouter/templates/gorouter.yml.erb
+++ b/jobs/gorouter/templates/gorouter.yml.erb
@@ -28,13 +28,6 @@ def parse_ip (ip, var_name)
   end
 end
 
-def validate_max_header_kb (kb)
-  if kb < 1 or kb > 1024
-    raise "Invalid router.max_header_kb value. Must be between 1 and 1,024 kb"
-  end
-  kb * 1024
-end
-
 def validate_balancing_algorithm (algorithm)
   valid_balancing_algorithms = ['round-robin', 'least-connection']
   unless valid_balancing_algorithms.include?(algorithm)
@@ -82,6 +75,27 @@ def status_tls
   return tls
 end
 
+def max_request_header_bytes
+  # See property description for details.
+  if_p("router.max_request_header_kb") do |kb|
+    return kb * 1024
+  end
+
+  kb = p("router.max_header_kb")
+  if kb < 1 or kb > 1024
+    raise "Invalid router.max_header_kb value. Must be between 1 and 1,024 kb"
+  end
+
+  return kb * 1024
+end
+
+def validate_header_count_limit (limit)
+  if limit > 0 && limit < 30
+    raise 'header limit too low, should be disabled (0) or at least 30'
+  end
+
+  return limit
+end
 
 params = {
   'zone' => spec.az,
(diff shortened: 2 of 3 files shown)


jobs/gorouter/spec
```diff
@@ -266,12 +266,41 @@ properties:
...
+        existing setups. If you previously configured router.max_header_kb it is recommended to
+        switch to this property instead.
+    example: 1024 # 1Mb
+  router.max_response_header_kb:
+    description: |
+        This value controls the maximum number of bytes (in KB) the gorouter will read
+        parsing the response header's keys and values, including the request
+        line. It does not limit the size of the response body. Responses with
+        larger headers will result in a 502 status code. A limit of zero or less will
+        result in the default GoLang limit being used.
```
peanball, line 288: do we know what that limit is? From the code I remember seeing 1MB.

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -145,7 +151,10 @@ params = {
   'route_services_internal_server_port' => p('router.route_services_internal_server_port'),
   'route_services_hairpinning_allowlist' => p('router.route_services_internal_lookup_allowlist'),
   'extra_headers_to_log' => p('router.extra_headers_to_log'),
-  'max_header_bytes' => validate_max_header_kb(p('router.max_header_kb')),
+  'max_request_header_bytes' => max_request_header_bytes,
+  'max_response_header_bytes' => p('router.max_response_header_kb') * 1024,
+  'max_request_headers' => p('router.max_request_headers'),
+  'max_response_headers' => p('router.max_response_headers'),
```
peanball: could you add spec tests for these properties also? yes, they're just passed but still we have tests for other properties.

jobs/gorouter/spec
maxmoehl: No that limit is different, as we don't control it I decided to not document it.

jobs/gorouter/templates/gorouter.yml.erb
maxmoehl: done.

peanball (dismissed):

jobs/gorouter/spec
```diff
@@ -266,12 +266,41 @@ properties:
       description: "Enforce strict validation of a route service signature"
       default: false
   router.max_header_kb:
+    description: |
+        Deprecated, use router.max_request_header_kb instead which is equivalent to this option.
+    default: 1024 # 1Mb
+  router.max_request_header_kb:
     description: |
         This value controls the maximum number of bytes (in KB) the gorouter will read
```
ameowlia, line 274: You have two descriptions here. I think you meant to combine them?

ameowlia (changes requested):

jobs/gorouter/spec
maxmoehl: The one for the existing property `max_header_kb` got cut to the deprecation notice and I've added `max_request_header_kb` with all the details. The diff is a bit weird because it noticed that I kept most of the existing description from the previous option.

ameowlia:
## acceptance testing
I tested all of the properties as a black box and they all worked as expected.
| Property             | Set Value | Test | Result | Is this what we want? |
| :---------------- | :------: | ----: | ----: | ----: |
| router.max_response_header_kb  && router.max_response_headers     |   default (0, no limit)   | 100,000 additional response headers for a total of ~2.5M  headers |  200 OK | ✅  |
| router.max_response_header_kb   |   10   | 100,000 additional response headers for a total of ~2.5M  headers | 502 `< X-Cf-Routererror: endpoint_failure (net/http: HTTP/1.x transport connection broken: net/http: server response headers exceeded 10240 bytes; aborted)`  | ✅  |
| router.max_response_header_kb    |  10   | 100 additional response headers for a total of ~2.8Kb  headers | 200 OK  | ✅ |
| router.max_response_headers    |   10   | 10 additional response headers | 502 `X-Cf-Routererror: endpoint_failure (too many response headers)`  | ✅|
| router.max_response_headers     |  10   | 8 additional response headers  | 200 OK  | ✅ |
| router.max_header_kb (this only ever affected the request headers)    |  default (1024 # 1Mb)   | 100,000 additional request headers for a total of ~2.5M  headers | 431 Request Header Fields Too Large  | ✅ |
| router.max_header_kb (this only ever affected the request headers)    |  2   | 100 additional request headers for a total of ~2.8Kb  headers | 431 x_cf_routererror:"max-request-size-exceeded"  |  ✅ |
| router.max_header_kb (this only ever affected the request headers)    |  2   | 50 additional request headers for a total of ~1.4Kb  headers | 200 OK  | ✅ |
| router.max_header_kb  = 2, router.max_request_header_kb = 10  (the 10 should overwrite the 2) |  -   | 100 additional request headers for a total of ~2.8Kb  headers | 200 OK  | ✅ |
| router.max_request_header_kb  |  10  | 1000 additional request headers for a total of ~28Kb  headers | 431 x_cf_routererror:"max-request-size-exceeded"  | ✅ |
| router.max_request_headers  |  100  | 80 additional request headers  | 200 OK | ✅ |
| router.max_request_headers  |  100  | 100 additional request headers | 431 x_cf_routererror:"max-request-size-exceeded"  | ✅ |


  ## new proxy handlers
I wrote 2 new proxy handlers to test this code.
* [Adding extra response headers](https://github.com/cloudfoundry/cf-networking-release/pull/310/files)
* [Adding extra request headers](https://github.com/cloudfoundry/cf-networking-release/blob/extra-request-headers/src/example-apps/proxy/handlers/extra_request_headers.go). This one is still WIP.

ameowlia (changes requested):
## Suggestion: set a minimum for the max_response/request_headers.
While I was testing this I set both max_response_headers and max_request_headers to 10. Silly me. This breaks basically every cf cli command. Maybe you could set a minimum to prevent people from doing dumb things like me? 20? 30?

```
# result when I set max_response_headers to 10

$ cf apps
Getting apps in org system / space system as admin...

Unexpected Response
Response Code: 502
Code: 0, Title: , Detail: 502 Bad Gateway: Registered endpoint failed to handle the request.

FAILED
```


## Check tests
When I ran all of the tests I got a failure for [this test about max header bytes](https://github.com/cloudfoundry/gorouter/blob/9ee77f4fdebd038c1978ecd5982aa2d3ae88b503/config/config_test.go#L214-L223).

ameowlia:
I still haven't done a thorough review of the code itself. I was busy today writing new handlers for the proxy app to make this acceptance testing easier. I'll get back to looking at this on Monday.

maxmoehl:
Thanks for taking the time to review this PR so thoroughly!

> While I was testing this I set both max_response_headers and max_request_headers to 10. Silly me. This breaks basically every cf cli command. Maybe you could set a minimum to prevent people from doing dumb things like me? 20? 30?

That sounds reasonable, I'll go with 30 to be safe. The template will then either allow `0` for disabling the limit or `>=30` for enabling, anything in between would fail the templating.

> When I ran all of the tests I got a failure for [this test about max header bytes](https://github.com/cloudfoundry/gorouter/blob/9ee77f4fdebd038c1978ecd5982aa2d3ae88b503/config/config_test.go#L214-L223).

Good catch, I somehow missed that! Addressed it in my last [push](https://github.com/cloudfoundry/gorouter/compare/d3b8890e3aea90525f1ba17391b63deb22a2f552..557b9c366ffe4617dc2c29950c6cbbb5cf464f37).

ameowlia (approved):

PR #446, opened 2024-11-26
feat: support syslog message truncation

maxmoehl:

Summary
---------------
Allow the user to specify a limit at which messages sent via syslog are truncated.

Backward Compatibility
---------------
Breaking Change? **No**

See also: cloudfoundry/gorouter#411

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -313,6 +313,9 @@ properties:
   router.logging.syslog_tag:
     description: "Tag to use when writing syslog messages"
     default: "vcap.gorouter"
+  router.logging.syslog_message_limit:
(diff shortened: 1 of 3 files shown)


jobs/gorouter/spec
```diff
@@ -313,6 +313,9 @@ properties:
   router.logging.syslog_tag:
     description: "Tag to use when writing syslog messages"
     default: "vcap.gorouter"
+  router.logging.syslog_message_limit:
```
ameowlia, line 316: Can you add more info here about what "0" means and the units that this limit is in?

ameowlia (approved):

ameowlia:
Reverted because this PR failed: https://github.com/cloudfoundry/gorouter/pull/411

PR #449, opened 2024-12-18
refactor: rename lb_algo to loadbalancing

a18e:

Summary
---------------
Renaming the `lb_algo` property to `loadbalancing` to align with cloud controller.
The per-route options feature ([RFC 0027](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0027-generic-per-route-features.md)) is not live yet, so this is not a breaking change.
See https://github.com/cloudfoundry/community/pull/1039 for an overview of all related PRs

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/jobs/route_registrar/spec
+++ b/jobs/route_registrar/spec
@@ -144,7 +144,7 @@ properties:
         writable (optional, boolean): sets the writable flag. defaults to false
 
       options object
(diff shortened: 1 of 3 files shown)


ameowlia:
⚠️ This alters bosh properties. We will need to mark this as a breaking change.

ameowlia (approved):

PR #450, opened 2025-01-13
test: fix acceptance tests on Apple Silicon Macs

peanball:
We observed sporadic compilation or bytecode issues
on Apple Silicon Macs.

According to https://github.com/golang/go/issues/42774,
the env variable `GODEBUG=asyncpreemptoff=1` works around
the issue by disabling asynchronous preemption of
goroutines.

This might impact garbage collection performance, but restores
the same level of reliability to locally running tests as
when run in Docker on an Intel based Mac.

The `GODEBUG=asyncpreemptoff=1` variable is set when creating
the Docker container for running tests locally, and applies
to all Go code, including `go vet`, for which we also observed
sporadic issues.

### DIFF
--- a/scripts/create-docker-container.bash
+++ b/scripts/create-docker-container.bash
@@ -34,12 +34,20 @@ else
   ARGS="${*}"
 fi
 
+# disable asyncpreempt on MacOS ARM64 (Apple Silicon) as it may cause sporadic errors in emulation
+# see: https://github.com/golang/go/issues/42774
+DISABLE_ASYNC_PREEMPT=
+if [ "$(uname -sm)" == "Darwin arm64" ]; then
+    DISABLE_ASYNC_PREEMPT='--env GODEBUG=asyncpreemptoff=1'
+fi
+
 docker pull "${IMAGE}"
 docker rm -f $CONTAINER_NAME
 docker run -it \
   --env "DB=${DB}" \
   --env "REPO_NAME=$REPO_NAME" \
   --env "REPO_PATH=/repo" \
+  $DISABLE_ASYNC_PREEMPT \
   --rm \
   --name "$CONTAINER_NAME" \
   -v "${REPO_PATH}:/repo" \
@@ -47,4 +55,3 @@ docker run -it \
   ${ARGS} \
   "${IMAGE}" \
   /bin/bash
-  


scripts/create-docker-container.bash
```diff
@@ -34,17 +34,24 @@ else
   ARGS="${*}"
 fi
 
+# disable asyncpreempt on MacOS ARM64 (Apple Silicon) as it may cause sporadic errors in emulation
+# see: https://github.com/golang/go/issues/42774
+DISABLE_ASYNC_PREEMPT=
+if [ "$(uname -s) $(uname -m)" == "Darwin arm64" ]; then
```
maxmoehl:
```suggestion
if [ "$(uname -sm)" == "Darwin arm64" ]; then
```

maxmoehl (dismissed):

scripts/create-docker-container.bash
peanball: applied, and TIL. Thanks!

ameowlia:
❓ Can you show the commands you ran and the errors you got before hand?

peanball:
> ❓ Can you show the commands you ran and the errors you got before hand?

Command (directly on the Mac, with Docker for Mac installed):
```shell
cd routing-release
scripts/test-in-docker.bash
```

mostly it's lines ending in `signal: illegal instruction`, as mentioned in https://github.com/golang/go/issues/42774.

```plain
...
[1736868869] Registry Suite - 88/88 specs - 7 procs •••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••• SUCCESS! 5.131790127s
Failed to compile container:

code.cloudfoundry.org/gorouter/registry/container_test: /usr/local/go/pkg/tool/linux_amd64/vet: signal: illegal instruction

...
Ginkgo ran 30 suites in 5m31.781438611s

There were failures detected in the following suites:
  container ./registry/container [Compilation failure]

...
```

peanball:
ok, so I reran it a couple of times now and occasionally still get the illegal instruction, even with the flag set. I will put it back to draft and check some more things.

It is possible that the env variable is not having an effect in all places, or it just doesn't help.

maxmoehl:
I'll go ahead and close this for now. If you manage to find a fix please open this again or create a new PR.

PR #454, opened 2025-02-08
Inline multierror sub-module

maxmoehl:

Summary
---------------

~Instead of using submodules for a repository which is [only referenced in a single release](https://github.com/search?q=org%3Acloudfoundry%20%22src%2Fcode.cloudfoundry.org%2Fmultierror%22&type=code) we should inline the code to make maintaining it easier.~ It seems to be used by [other orgs](https://github.com/search?q=%22https%3A%2F%2Fgithub.com%2Fcloudfoundry%2Fmultierror%22+path%3A.gitmodules&type=code). This will reduce overhead by removing the need to:
* bump submodules
* trace changes across multiple repositories
* maintain multiple repositories
* ...

For repositories referenced multiple times this process most probably does not apply (there are [ways](https://man.archlinux.org/man/git-subtree.1) to do it but I'm not sure I like them), but those seem to be rare.

I'm opening this as a draft PR to give room for discussion. To inline a repository, first remove the sub-module (feadf3a33a48be857a0bcd754df96b693d45f20f) and then inline the code via `git subtree add --prefix src/code.cloudfoundry.org/multierror https://github.com/cloudfoundry/multierror main`

The easiest way to do this is to enforce a brief development-freeze on the repository which should be inlined (multierror doesn't seem to have active development at the moment). Branches could be transferred as well but it's easier to freeze development for a few days, especially since our repos are not highly active.

What I don't know is the changes we would have to make to our CI systems.

/cc @ameowlia @peanball with whom I discussed this previously.

Backward Compatibility
---------------
Breaking Change? **No**


maxmoehl:
(sorry for the pings, I wasn't expecting the CLA bot to pick up on draft PRs)

ameowlia:
Yesssss. I love this. 


* ❓ What did you do to confirm that no other release is using this submodule? 
* I don't think this will break CI. The submodules are bumped by calling `git submodule sync --recursive`. Given that it will no longer be a submodule, I think everything will continue working properly.
* I think there are more files we can delete (license, PR request template, gitignore, etc)

maxmoehl:
> * ❓ What did you do to confirm that no other release is using this submodule?

I'm assuming that every repository which consumes this submodule does so via the `.gitmodules` file, you can easily search that using [this](https://github.com/search?q=org%3Acloudfoundry%20%22src%2Fcode.cloudfoundry.org%2Fmultierror%22&type=code) query. Widening the query to [all of GitHub](https://github.com/search?q=%22https%3A%2F%2Fgithub.com%2Fcloudfoundry%2Fmultierror%22+path%3A.gitmodules&type=code) and adjusting to filter better reveals quite some copies (?) of the routing-release as well as non-CF consumers.

So this might not be the best candidate after all... Do we have a policy on such things?

> * I don't think this will break CI. The submodules are bumped by calling `git submodule sync --recursive`. Given that it will no longer be a submodule, I think everything will continue working properly.

:+1:

> * I think there are more files we can delete (license, PR request template, gitignore, etc)

I will probably look into doing a different module first and will add a commit to clean it up.

ameowlia:
Re all the non-CF consumers: 
* Many are routing-release forks. I am not worried about these.
* Many are bosh bootloaders forkers. The [main repo of bosh bootloader](https://github.com/cloudfoundry/bosh-bootloader/blob/main/.gitmodules) does not include this submodule anymore, so I am not worried about these forks.
* The log-search-for-cf ones havent been updated for 6 or 7 years respectively. So I am not worried about those.

Given that, I still think this is a good candidate.

geofffranks:
FYI - https://pkg.go.dev/code.cloudfoundry.org/multierror?tab=importedby is a good way to check that nothing outside of cloudfoundry is importing these packages before we make them un-importable.

maxmoehl:
> FYI - https://pkg.go.dev/code.cloudfoundry.org/multierror?tab=importedby is a good way to check that nothing outside of cloudfoundry is importing these packages before we make them un-importable.

Good idea!

We should probably look into how `code.cloudfoundry.org` resolves packages. With some magic we might be able to still serve those packages. After all, `github.com/cloudfoundry/routing-release/src/code.cloudfoundry.org/multierror` would also be valid import-path (except that it's invalid at the moment because the module declares itself as `code.cloudfoundry.org`). @geofffranks do you know who is familiar with that vanity-URL-resolver?

geofffranks:
Unfortunately I do not.

maxmoehl:
I asked @beyhan to ask about the `code.cloudfoundry.org` thing in the TOC meeting.

beyhan:
@maxmoehl the right place to adjust the rules is the https://github.com/cloudfoundry/go-fetcher by opening a pr.

maxmoehl:
The discussion moved to Slack: https://cloudfoundry.slack.com/archives/C02HNDJB31R/p1739364912266279

TL;DR: there is no easy fix to keep inlined packages import-able.

PR #455, opened 2025-02-12
refactor: move timing info to extra access log fields logic

maxmoehl:

Summary
---------------
Now that we have to ability to add individual log fields I've migrated a set of log fields that was previously behind a special feature flag to use the same logic internally. The feature flag remains but gorouter is no longer aware of it.

The output is identical because the logic for the attempts logging and the dynamic logging is right next to each other. The template ensures that the order of the attempts logging fields is the same as it was previously in the gorouter code.

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -328,7 +328,7 @@ properties:
       this property and operators have to explicitly enable them. This is done to prevent breaking
       log parsing in existing setups by the introduction of new fields. Does not affect stdout /
       stderr logs.
-      Available fields are: local_address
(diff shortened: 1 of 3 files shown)


ameowlia (commented):
Thank you for making it backwards compatible!

peanball (approved):
Tests executed, code reviewed, functionality tested in a dev environment. All working.

peanball:
gorouter has been bumped in routing release. I'm merging this PR now.

PR #456, opened 2025-02-12
feat: Introduce switch for envelope v1 metrics

b1tamara:

Summary
---------------
- Implement a switch to enable / disable metrics reported using Envelope V1
- Introduce a new gorouter prometheus property `enabled` which has been automatically set to true if prometheus.port is set. 
- Do not exclude metrics reporting by prometheus and using Envelope V1.
- Adapt `prom_scraper` configuration template to use https mode only if the certificates have been set. 
- Introduce a new flag to enable / disable `prom_scraper` for the use cases in which the metrics will be read by another consumers (e.g. telegraf). By default it should be enabled. 

Backward Compatibility
---------------
Breaking Change? **No**

Complementing Gorouter PR: https://github.com/cloudfoundry/gorouter/pull/466

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -76,8 +76,11 @@ properties:
     description: "TLS Certificate used for the TLS listener of the LB healthcheck endpoint"
   router.status.tls.key:
     description: "Private Key used for the TLS listener of the LB healthcheck endpoint"
+  router.enable_envelope_v1_metrics:
+      description: "Enables support for metrics reported via Envelope."
+      default: true
   router.prometheus.port:
-    description: "Port for the prometheus endpoint."
+    description: "Port for the prometheus endpoint, automatically enables Prometheus support."
   router.prometheus.server_name:
     description: "The server name used in the certificate for the metrics endpoint."
   router.prometheus.ca_cert:
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -93,13 +93,13 @@ properties:
     description: "TLS private key for prometheus server."
     default: ""
   router.prometheus.meters.route_lookup_time_histogram_buckets:
-    description: "Upper limits of the ranges in which the observed value of route lookup time is expected to fall"
+    description: "Upper limits in nanoseconds of the ranges in which the observed value of route lookup time is expected to fall"
     default: [10000, 20000, 30000, 40000, 50000, 60000, 70000, 80000, 90000, 100000]
   router.prometheus.meters.route_registration_latency_histogram_buckets:
-    description: "Upper limits of the ranges in which the observed value of route registration latency is expected to fall"
+    description: "Upper limits in milliseconds of the ranges in which the observed value of route registration latency is expected to fall"
     default: [0.1, 0.5, 1, 1.5, 2, 2.5, 3, 3.5, 4]
   router.prometheus.meters.routing_response_latency_histogram_buckets:
-    description: "Upper limits of the ranges in which the observed value of route response latency is expected to fall"
+    description: "Upper limits in milliseconds of the ranges in which the observed value of route response latency is expected to fall"
     default: [1, 2, 4, 6, 8, 10, 20, 40, 50, 100, 500, 1000]
   router.requested_route_registration_interval_in_seconds:
     description: |
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -80,7 +80,7 @@ properties:
       description: "Enables support for metrics reported via Envelope."
       default: true
   router.prometheus.port:
-    description: "Port for the prometheus endpoint, automatically enables Prometheus support."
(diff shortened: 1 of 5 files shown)


ctlong:
If you want the prometheus metrics to be scraped by the Prom Scraper agent, then you'll need to configure a `prom_scraper_config.yml`. You can read more about that [here](https://github.com/cloudfoundry/loggregator-agent-release/blob/main/docs/prom-scraper.md).

mike-jc:
@ctlong There's already a template for scraper config - https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/templates/prom_scraper_config.yml.erb. I believe that is what you are talking about and that should be enough, right?

chombium:
@ctlong I've checked the file which @mike-jc has linked, that's enough to configure the Prom Scraper.

ctlong:
I agree!

I didn't check for an existing prom scraper config because I assumed that there was no Prometheus server in Gorouter previously. Is there already a server or was this config telling Prom Scraper to scrape a thing that did not exist?

peanball:
@ctlong, there was a Prometheus metrics experiment in Gorouter some years ago. That came with a prom scraper config. We're now fleshing out the experiment to a full feature.

geofffranks:
I took a cursory look at this, and it looks good. I don't see any changes that would end up breaking things without operators opting in, or causing issues with edge case configurations.

This will still need a full review from an ARP approver, running tests, ensuring deployment works as intended, and validation of doc updates for the new prometheus metrics.

geofffranks:
Thanks for making envelope the default and allowing opt-in to prometheus + opt-out for envelope!

jobs/gorouter/spec
```diff
@@ -89,6 +92,15 @@ properties:
   router.prometheus.key:
     description: "TLS private key for prometheus server."
     default: ""
+  router.prometheus.meters.route_lookup_time_histogram_buckets:
+    description: "Upper limits of the ranges in which the observed value of route lookup time is expected to fall"
```
peanball: what's the time unit here?

jobs/gorouter/spec
```diff
@@ -76,8 +76,11 @@ properties:
     description: "TLS Certificate used for the TLS listener of the LB healthcheck endpoint"
   router.status.tls.key:
     description: "Private Key used for the TLS listener of the LB healthcheck endpoint"
+  router.enable_envelope_v1_metrics:
+      description: "Enables support for metrics reported via Envelope."
+      default: true
   router.prometheus.port:
-    description: "Port for the prometheus endpoint."
+    description: "Port for the prometheus endpoint, automatically enables Prometheus support."
```
peanball:
```suggestion
    description: "Port for the prometheus endpoint. Automatically enables Prometheus support."
```

jobs/gorouter/spec
```diff
@@ -89,6 +92,15 @@ properties:
   router.prometheus.key:
     description: "TLS private key for prometheus server."
     default: ""
+  router.prometheus.meters.route_lookup_time_histogram_buckets:
+    description: "Upper limits of the ranges in which the observed value of route lookup time is expected to fall"
+    default: [10000, 20000, 30000, 40000, 50000, 60000, 70000, 80000, 90000, 100000]
+  router.prometheus.meters.route_registration_latency_histogram_buckets:
+    description: "Upper limits of the ranges in which the observed value of route registration latency is expected to fall"
```
peanball: Time unit would be useful again in the description

jobs/gorouter/spec
```diff
@@ -89,6 +92,15 @@ properties:
...
+    default: [0.1, 0.5, 1, 1.5, 2, 2.5, 3, 3.5, 4]
+  router.prometheus.meters.routing_response_latency_histogram_buckets:
+    description: "Upper limits of the ranges in which the observed value of route response latency is expected to fall"
```
peanball: Time unit would be useful again in the description

peanball (commented):
The time units for new values would be nice, at least in the description. It's probably milliseconds, but without looking at the impl, we don't know.

jobs/gorouter/spec
mike-jc: Good call. Specified unit.

jobs/gorouter/spec
mike-jc: Good call. I've specified unit now.

jobs/gorouter/spec
mike-jc: Good call. I've specified unit now.

peanball:
Verified that metrics are created on the prometheus endpoint and scraped by the prom_scraper in a test environment.

peanball (approved):

peanball:
Bumping gorouter submodule via https://ci.funtime.lol/teams/wg-arp-networking/pipelines/routing-release/jobs/bump-dependencies-go-mod/builds/285

peanball:
acceptance tests in docker re-run successfully. Merging.

PR #461, opened 2025-04-10
dep: bump gopkg.in/yaml to v3 for nats_client

mtekel:
Summary
---------------
Bump yaml gopkg to v3 for nats_client.

Backward Compatibility
---------------
Breaking Change? No


maxmoehl (commented):
Code-wise this looks good as we only parse very basic things here, have you run the nats_client manually after changing the dependency? It seems like there are no tests for it.

maxmoehl (approved):
I've thrown it onto one of our dev envs and it works fine. Since it parses the gorouter config and gorouter is also switched to v3 already I don't see why this shouldn't work.

PR #462, opened 2025-04-11
ci: make docker scripts podman-compatible

maxmoehl:

Summary
---------------
Podman is mostly compatible with the docker CLI except for a few small differences. This commit alignes the docker scripts so that it is compatible with both.

By default podman requires fully qualified image names and the `rm` command does not ignore errors when removing a containter that does not exist so we simply `|| true` after it.

Backward Compatibility
---------------
Breaking Change? **No**


mtekel (approved):
works fine with docker

plowin (approved):
lgtm

peanball:
This also works fine when pointing the docker client (e.g. via context) to the Podman socket, with or without these changes.

PR #464, opened 2025-04-11
fix: add OU in client-cert field_map

plowin:
see https://www.rfc-editor.org/rfc/rfc1779#section-2.3

We may consider removing ON independently, but decided to introduce OU additionally to avoid any side-effects and support a valid scenario.

Summary
---------------

Backward Compatibility
---------------
Breaking Change? **No**


plowin:
~~@ameowlia, maybe you can comment if you are aware of people using this feature or would be supporting just to do the change as proposed and add a small remark in the release notes~~

Edit: we rediscussed and introduce OU additionally. Even though ON is not defined in the RFC, it might be used and can be kept. This change cannot break existing scenarios as a cert with OU would have been rejected before.

peanball (approved):
Not sure where I had taken `organization_name` from anymore...

Thanks for the fix.

PR #471, opened 2025-04-30
include http/1 and http/2 request timeout values

kart2bc:

Summary
---------------

Backward Compatibility
---------------
Breaking Change? **Yes/No**

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -523,6 +523,16 @@ properties:
       Requests from router to backend endpoints that are longer than this duration will be canceled and logged as
       `backend-request-timeout` errors. If set to 0 this timeout is disabled.
     default: 900
+  http_1_request_timeout_in_seconds:
+    description: |
+      The amount of time HTTP/1 requests from the Gorouter to apps are allowed to live before being canceled. 
+      Enter value in seconds. Set to -1 to disable timeouts. If set to 0, it inherits the value from request_timeout_in_seconds.
+    default: 0
+  http_2_request_timeout_in_seconds:
+    description: |
+      The amount of time HTTP/2 requests from the Gorouter to apps are allowed to live before being canceled. 
+      Enter value in seconds. Set to -1 to disable timeouts. If set to 0, it inherits the value from request_timeout_in_seconds.
+    default: 0
   endpoint_dial_timeout_in_seconds:
     description: |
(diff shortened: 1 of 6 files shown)


src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -440,6 +440,8 @@ type Config struct {
 	PublishActiveAppsInterval       time.Duration `yaml:"publish_active_apps_interval,omitempty"`
 	StartResponseDelayInterval      time.Duration `yaml:"start_response_delay_interval,omitempty"`
 	EndpointTimeout                 time.Duration `yaml:"endpoint_timeout,omitempty"`
+	Http1EndpointTimeout            time.Duration `yaml:"endpoint_timeout,omitempty"`
+	Http2EndpointTimeout            time.Duration `yaml:"endpoint_timeout,omitempty"`
```
geofffranks: These struct tags will need to be updated + match what's referenced in the job template

kart2bc: updated the json tag values.

jobs/gorouter/spec
```diff
@@ -523,6 +523,18 @@ properties:
...
+    description: |
+      This configures an entire request timeout.
+      HTTP/1 requests from router to backend endpoints that are longer than this duration will be canceled and logged as
+      `backend-request-timeout` errors. If set to 0 this timeout is disabled.
+    default: 900
+  http_2_request_timeout_in_seconds:
+    description: |
+      This configures an entire request timeout.
+      HTTP/2 requests from router to backend endpoints that are longer than this duration will be canceled and logged as
+      `backend-request-timeout` errors. If set to 0 this timeout is disabled.
```
geofffranks: These should say something like "if set to -1, this timeout is disabled. If set to 0, it inherits the value from `request_timeout_in_seconds`"

geofffranks (changes requested):

jobs/gorouter/spec
maxmoehl: Couldn't we use set / unset to decide whether to use `request_timeout_in_seconds` or `http_N_request_timeout_in_seconds`? Then we can avoid the additional magic value.

geofffranks (approved):

Issue #408, opened 2024-05-08
Allow Route Services to be used with Web Sockets


domdom82:
### Proposed Change

**As a** developer
**I want** to attach a route service to my web socket app
**So that** I can rate-limit it using the route service

We currently prevent web socket upgrades on routes that have a route service attached.

The check is implemented [here](https://github.com/cloudfoundry/gorouter/blob/main/handlers/routeservice.go#L81-L91)

However, there is no technical reason why route services should not be able to handle web socket requests. I've tested this by commenting out the above check and attaching a route service to a web socket app. The handshake and traffic work just fine through the route service.

So long as the route service behaves correctly as a [reverse proxy](https://pkg.go.dev/net/http/httputil#ReverseProxy) and forwards the `Connection: Upgrade` and `Upgrade: Websocket` headers to the destination app, there is no reason to block it at gorouter. After all, web socket requests are just HTTP requests.

This issue has been brought up by one of our customers who need a route service to rate-limit their app and have been complaining that it didn't work with web socket apps.

**Note:** The check is very old. It might be outdated. We recently [removed the RequestHandler](https://github.com/cloudfoundry/gorouter/pull/403) and let Go do all the work for web sockets. My suspicion is that the check was added because RequestHandler didn't play well together with route services.

### Acceptance criteria

**Scenario**: My web socket app needs rate limiting via a route service.
**Given**
```
cf push my-websocket-app
cf push ratelimiter
cf create-user-provided-service ratelimiter-service -r https://ratelimiter.bosh-lite.com
cf bind-route-service bosh-lite.com ratelimiter-service --hostname my-websocket-app
```
**When** I connect to wss://my-websocket-app.bosh-lite.com using a web socket client
**Then** The connection will work through the route service as expected

### Related links

- [removal of legacy RequestHandler](https://github.com/cloudfoundry/gorouter/pull/403)
- [sample rate-limiter](https://github.com/cloudfoundry-samples/ratelimit-service)

maxmoehl:
Some more details:

* The check was introduced with 1bf9a13a98ef2230760b572186d52acc0fad9bad.
* Broadcom shut down the Pivotal tracker and it seems like #149929938 didn't make it into any of archives :,(

We are planning to remove the check eventually as our customers have use-cases for such scenarios and it doesn't seem like there is a good reason not to allow it.

maxmoehl:
[...]

PR #474, opened 2025-05-05
WebSocket Support for Apps with Route-Services

Dariquest:
Summary
---------------

Route services should be able to handle web socket requests. 

The Gorouter flag "route_services.enable_websockets" is introduced to enable/disable web socket communication for apps with route services. The default value is true.

```
 router.route_services.enable_websockets:
    description: Enable websocket connections for application routes bound to Route Services.
    default: true
```

Web socket upgrades on routes with an attached route service were prevented previously. 
This issue has been brought up by one of our customers, who needed a route service to rate-limit their app.

[GitHub Issue](https://github.com/cloudfoundry/routing-release/issues/408)

Backward Compatibility
---------------
Breaking Change? **No**

Test
---------------
**Scenario**: A web socket app needs rate limiting via a route service.
**Given**

```
cf push websocket-app
cf push ratelimiter
cf create-user-provided-service ratelimiter-service -r https://ratelimiter.cfapps.xxx.sapcloud.io
cf bind-route-service cfapps.xxx.sapcloud.io ratelimiter-service --hostname websocket-app
```

**When** I connect to wss://websocket-app.cfapps.xxx.sapcloud.io using a web socket client or simply curl it with the corresponding websocket request headers
**Then** The connection will work through the route service as expected

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -255,6 +255,10 @@ properties:
       Maximum number of attempts on failing requests against route service URLs.
       The minimum value for this setting is 1. This prevents gorouter from getting blocked by indefinite retries.
     default: 3
+  router.route_services.enable_websockets:
+    description: |
+      Enable websocket connections for application routes bound to Route Services.
+    default: true
   router.route_services.cert_chain:
     description: Certificate chain used for client authentication to TLS-registered route services.  In PEM format.
   router.route_services.private_key:
--- a/src/code.cloudfoundry.org/gorouter/integration/route_services_test.go
+++ b/src/code.cloudfoundry.org/gorouter/integration/route_services_test.go
@@ -4,29 +4,52 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io"
+	"log"
+	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/http/httputil"
+	"net/url"
+	"os"
 	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"golang.org/x/net/websocket"
 
 	"code.cloudfoundry.org/gorouter/route"
 	"code.cloudfoundry.org/gorouter/test/common"
 )
+
+func wsClient(conn net.Conn, urlStr string) (*websocket.Conn, error) {
+	wsUrl, err := url.ParseRequestURI(urlStr)
+	Expect(err).NotTo(HaveOccurred())
+
+	cfg := &websocket.Config{
+		Location: wsUrl,
+		Origin:   wsUrl,
+		Version:  websocket.ProtocolVersionHybi13,
+	}
+
+	wsConn, err := websocket.NewClient(cfg, conn)
+	return wsConn, err
+}
 
 var _ = Describe("Route services", func() {
 
 	var testState *testState
 
 	const (
-		appHostname = "app-with-route-service.some.domain"
+		appHostname   = "app-with-route-service.some.domain"
+		wsAppHostname = "ws-app-with-route-service.some.domain"
 	)
 
 	var (
-		testApp      *httptest.Server
-		routeService *httptest.Server
+		testApp        *httptest.Server
+		routeService   *httptest.Server
+		wsTestApp      *httptest.Server
+		wsRouteService *httptest.Server
 	)
 
 	BeforeEach(func() {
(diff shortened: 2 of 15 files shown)


Dariquest:
Performed a test of this Routing-release Gorouter change on a dev landscape with a sample websocket-app to show that the apps with bound route services support web sockets:

```
cf push websocket-app
cf push ratelimiter
cf create-user-provided-service ratelimiter-service -r https://ratelimiter.cfapps.xxx.sapcloud.io
cf bind-route-service cfapps.xxx.sapcloud.io ratelimiter-service --hostname websocket-app
```
1. Using a web socket client wss:

wscat -c wss://websocket-app.cfapps.xxx.sapcloud.io/ws
Connected (press CTRL+C to quit)
> test
< test

2. Curling the sample app with the corresponding websocket connection headers:

```
$ curl -i -N -H "Connection: Upgrade"      -H "Upgrade: websocket"      -H "Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ=="      -H "Sec-WebSocket-Version: 13"      -H "Origin: https://websocket-app.cfapps.xxx.sapcloud.io"      https://websocket-app.cfapps.xxx.sapcloud.io/ws --http1.1
HTTP/1.1 101 Switching Protocols
connection: Upgrade
sec-websocket-accept: xxx
upgrade: websocket
x-vcap-request-id: 0237d2fc-72f6-4905-71c3-88a2419fd40e
strict-transport-security: max-age=31536000; includeSubDomains; preload;
```

src/code.cloudfoundry.org/gorouter/integration/route_services_test.go
```diff
@@ -102,7 +209,7 @@ var _ = Describe("Route services", func() {
 				req := testState.newGetRequest(
 					fmt.Sprintf("https://%s", appHostname),
 				)
-
+				fmt.Printf("hostname: %s", appHostname)
```
maxmoehl: Probably a left-over from debugging the tests?

src/code.cloudfoundry.org/gorouter/integration/route_services_test.go
```diff
@@ -69,6 +92,30 @@ var _ = Describe("Route services", func() {
...
+					req.URL = &url.URL{
+						Scheme: "http",
+						Host:   fmt.Sprintf("127.0.0.1:%d", testState.cfg.Port),
+					}
+					req.Host = forwardedURL.Host
+				},
+				Transport: &http.Transport{
+					TLSClientConfig: &tls.Config{
+						MinVersion:         tls.VersionTLS12,
+						InsecureSkipVerify: true, //nolint:gosec // required, as route-service needs to talk to crt-test, which does not have a valid certificate
```
maxmoehl: I think this comment got copied from somewhere else, please remove it. If it fails because of TLS errors we will have to check where to get the correct TLS config from.

src/code.cloudfoundry.org/gorouter/integration/route_services_test.go
```diff
@@ -84,6 +131,66 @@ var _ = Describe("Route services", func() {
...
+					testState.cfg.SSLPort,
+				)
+			})
+			AfterEach(func() {
+				f, err := os.Open(accessLogFile)
+				Expect(err).NotTo(HaveOccurred())
+
+				_, err = io.Copy(os.Stdout, f)
+				Expect(err).NotTo(HaveOccurred())
+			})
```
maxmoehl: Please remove this block, it was only intended for debugging.

src/code.cloudfoundry.org/gorouter/integration/route_services_test.go
Dariquest: yes, thanks!

jobs/gorouter/spec
```diff
@@ -255,6 +255,10 @@ properties:
       Maximum number of attempts on failing requests against route service URLs.
       The minimum value for this setting is 1. This prevents gorouter from getting blocked by indefinite retries.
     default: 3
+  router.route_services.disable_websockets:
+    description: |
+      Disable websocket connections for application routes bound to Route Services.
+    default: false
```
geofffranks: Can you default this to `true` please? 

It looks like we explicitly disabled websockets for route-services back in 2017 https://github.com/cloudfoundry/routing-release/commit/1bf9a13a98ef2230760b572186d52acc0fad9bad, so I don't want to surprise users without opting into a change in behavior like this.

geofffranks (changes requested):
Requested default change

jobs/gorouter/spec
Dariquest: Hi @geofffranks, I have changed the default value and adjusted the tests. The flag should be in this case imho better  renamed to "enable_webockets" with the default value of "false". And this renaming brings me to the thought, that it makes more sense to enable web sockets by default and disable them in emergency case. Why do one need a separate configuration step to enable something which is supposed to be supported by standard? It was my understanding from the corresponding GitHub Issue. I also understand that we want to keep the default functionality.

geofffranks: Thanks! I agree the naming is awkward. 

Our concern with enabling this by default was that it's been this way for years, and developers might run into problems with it suddenly changing behavior. Howeve this would only affect apps with websocket requests and route services.  It seems a little farfetched that someone would be relying on the current behavior of a 503 being returned constantly (and not adjust their app/route service).

I think you're right and it should be `router.route_services.enable_websockets` and `true` by default.

maxmoehl:
GitHub seems confused about workflows being there while also not being there on this branch. @Dariquest can you rebase once more?

geofffranks:
> GitHub seems confused about workflows being there while also not being there on this branch. @Dariquest can you rebase once more?

Everything looks good to me. Once the rebasing + GH actions are sorted out feel free to approve + merge.n

maxmoehl:
The workflow uses secrets, e.g. `${{ secrets.GCP_SERVICE_ACCOUNT_TAS_RUNTIME_BUILD_IMAGE_READER }}` and those are inaccessible if your PR is coming from a fork due to security policies.

We will ignore the validation for now but can have a discussion on how to fix this for future contributions.

/fyi @geofffranks @kart2bc

jobs/gorouter/spec
Dariquest: I have changed the naming and kept and reversed the default value 
  disable_websockets: false => enable_websockets: true

geofffranks (approved):

PR #476, opened 2025-05-05
Fix package spec more

mariash:

Summary
---------------
We need jq in the gorouter and had to manually add it

Backward Compatibility
---------------
Breaking Change? **No**


ebroberson (approved):
LGTM

PR #478, opened 2025-05-06
Route registry enhancements

### LINKED ISSUE #468: Fix Route Registry Metrics to Report Correct Result
### Proposed Change

We see an issue with the Route Registry code. 

#### Register

Currently, [Register()](https://github.com/cloudfoundry/gorouter/blob/b5be7589592b9cf21f69dd74055bdcc77157e369/registry/registry.go#L86-L86) calls [register()](https://github.com/cloudfoundry/gorouter/blob/b5be7589592b9cf21f69dd74055bdcc77157e369/registry/registry.go#L116-L116), which calls [Pool.Put()](https://github.com/cloudfoundry/gorouter/blob/b5be7589592b9cf21f69dd74055bdcc77157e369/route/pool.go#L276-L276)

`Put()` updates the pool if required, and returns the `PoolPutResult` to `register()`:
* UPDATED, whenever the endpoint has previously been registered in the pool (even if unmodified)
  * Exception: UNMODIFIED, if the message is a duplicate (based on ModificationTag)
* ADDED, if the endpoint doesn't exist in the pool yet 

`register()` creates the pool if required, sets the time of update and forwards the `PoolPutResult` to `Register()`. `Register()` receives and logs the `PoolPutResult`, and ingests a generic route registry metric.

#### Unregister

On the other hand, we have a similar functionality with [Unregister()](https://github.com/cloudfoundry/gorouter/blob/b5be7589592b9cf21f69dd74055bdcc77157e369/registry/registry.go#L167-L167), which calls [unregister()](https://github.com/cloudfoundry/gorouter/blob/b5be7589592b9cf21f69dd74055bdcc77157e369/registry/registry.go#L178-L178), which calls [pool.Remove()](https://github.com/cloudfoundry/gorouter/blob/b5be7589592b9cf21f69dd74055bdcc77157e369/route/pool.go#L365-L365).

However, the structure is quite different here:
`Remove()` returns a boolean value reflecting whether the route has been removed from the pool, or not. `unregister()` however does some logging based on the result of `Remove()`, and removes the pool if required. `Unregister()` finally unconditionally ingests a generic route unregistry metric.  

#### Issues

There are some problems with the current code:
* We have a very generic metric for registering and unregistering endpoints - it does not reveal the result of the (un)registration (endpoint added, modified, unchanged, etc). However, tests have shown that the cost for each case differs, and that it is beneficial for operations to have more detailed metrics.  
* `pool.Put()` always returns status `UPDATED` when the route exists in the pool. Hence, the metric doesn't reflect whether a route has received an update, or if not (in case its attributes haven't changed).
* `Unregister()` always ingests an `UnregistryMessage`, independently of the result of pool.Remove(). Hence, the metric doesn't reflect whether a route has been removed or not.
* We should introduce an additional metric that reflects whether a whole route has been created or removed upon `Register()` / `Unregister()`.
* The structure of the `Register()` and `Unregister()` functions and its related private functions should be refactored, e.g.
  * Code / function structure should be the same 
  * Logging should be done in the public function functionality
  * Improve code structure in general, make the long if-clauses in registry.go less complicated

### Acceptance criteria

* Have better metrics that reflect the actual result of registering and unregistering an endpoint, based on the `PoolPutResult` and a `PoolRemoveResult`.
  * Avoid breaking changes by using tags for the actual result in `CaptureRegistryMessage` and `CaptureUnregistryMessage`.
* Fix the wrong `UPDATED` result in case of no-op registry update. 
* Refactor the code to mitigate the issues described above.

### Related links

* Draft Pull-Request cloudfoundry/gorouter#456: This was created before being able to use metric tags (introduced with support for Prometheus based metrics). This Pull-Request already addresses the issues described here, but needs a final refactoring to have no breaking change due to metrics renaming.

maxmoehl:

Summary
---------------

Rework some of the route registry logic to reduce nested ifs and align some of the logic.

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/metrics/metricsreporter_test.go
+++ b/src/code.cloudfoundry.org/gorouter/metrics/metricsreporter_test.go
@@ -524,16 +524,16 @@ var _ = Describe("MetricsReporter", func() {
 		Expect(count).To(Equal(uint64(5)))
 	})
 
-	It("increments the routes_added metric", func() {
-		metricReporter.CaptureRouteAdded()
+	It("increments the routes_registered metric", func() {
+		metricReporter.CaptureRoutesRegistered()
 		Expect(batcher.BatchIncrementCounterCallCount()).To(Equal(1))
-		Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("routes_added"))
+		Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("routes_registered"))
 	})
 
-	It("increments the routes_deleted metric", func() {
-		metricReporter.CaptureRouteDeleted()
+	It("increments the routes_unregistered metric", func() {
+		metricReporter.CaptureRoutesUnregistered()
 		Expect(batcher.BatchIncrementCounterCallCount()).To(Equal(1))
-		Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("routes_deleted"))
+		Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("routes_unregistered"))
 	})
 
 	It("increments the backend_tls_handshake_failed metric", func() {
--- a/src/code.cloudfoundry.org/gorouter/metrics/metricsreporter_test.go
+++ b/src/code.cloudfoundry.org/gorouter/metrics/metricsreporter_test.go
@@ -467,7 +467,7 @@ var _ = Describe("MetricsReporter", func() {
 
 		It("sends number of nats messages received from each component", func() {
 			endpoint.Tags = map[string]string{}
-			metricReporter.CaptureRegistryMessage(endpoint, route.ADDED.String())
+			metricReporter.CaptureRegistryMessage(endpoint, route.EndpointAdded.String())
 
 			Expect(batcher.BatchIncrementCounterCallCount()).To(Equal(1))
 			Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("registry_message"))
@@ -475,10 +475,10 @@ var _ = Describe("MetricsReporter", func() {
 
 		It("sends number of nats messages received from each component", func() {
 			endpoint.Tags = map[string]string{"component": "uaa"}
-			metricReporter.CaptureRegistryMessage(endpoint, route.ADDED.String())
+			metricReporter.CaptureRegistryMessage(endpoint, route.EndpointAdded.String())
 
 			endpoint.Tags = map[string]string{"component": "route-emitter"}
-			metricReporter.CaptureRegistryMessage(endpoint, route.ADDED.String())
+			metricReporter.CaptureRegistryMessage(endpoint, route.EndpointAdded.String())
 
 			Expect(batcher.BatchIncrementCounterCallCount()).To(Equal(2))
 			Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("registry_message.uaa"))
--- a/src/code.cloudfoundry.org/gorouter/metrics/metricsreporter_test.go
+++ b/src/code.cloudfoundry.org/gorouter/metrics/metricsreporter_test.go
@@ -524,6 +524,18 @@ var _ = Describe("MetricsReporter", func() {
 		Expect(count).To(Equal(uint64(5)))
 	})
 
+	It("increments the routes_added metric", func() {
+		metricReporter.CaptureRouteAdded()
+		Expect(batcher.BatchIncrementCounterCallCount()).To(Equal(1))
+		Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("routes_added"))
+	})
+
+	It("increments the routes_deleted metric", func() {
+		metricReporter.CaptureRouteDeleted()
+		Expect(batcher.BatchIncrementCounterCallCount()).To(Equal(1))
+		Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("routes_deleted"))
+	})
+
 	It("increments the backend_tls_handshake_failed metric", func() {
 		metricReporter.CaptureBackendTLSHandshakeFailed()
 		Expect(batcher.BatchIncrementCounterCallCount()).To(Equal(1))
--- a/src/code.cloudfoundry.org/gorouter/registry/registry.go
+++ b/src/code.cloudfoundry.org/gorouter/registry/registry.go
@@ -88,12 +88,18 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
 		return
 	}
 
-	poolPutResult := r.register(uri, endpoint)
+	poolPutResult, routeAdded := r.register(uri, endpoint)
 
 	r.reporter.CaptureRegistryMessage(endpoint, poolPutResult.String())
 
 	if poolPutResult == route.EndpointAdded && !endpoint.UpdatedAt.IsZero() {
 		r.reporter.CaptureRouteRegistrationLatency(time.Since(endpoint.UpdatedAt))
+	}
+
+	if routeAdded {
+		r.logger.Info("route-registered", slog.Any("uri", uri))
+		// for backward compatibility:
+		r.logger.Debug("uri-added", slog.Any("uri", uri))
 	}
 
 	switch poolPutResult {
@@ -116,7 +122,7 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
 	}
 }
 
-func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.PoolPutResult {
+func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) (putResult route.PoolPutResult, routeAdded bool) {
 	r.RLock()
 	defer r.RUnlock()
 
@@ -127,7 +133,7 @@ func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.
 	if pool == nil {
 		// release read lock, insertRouteKey() will acquire a write lock.
 		r.RUnlock()
-		pool = r.insertRouteKey(routekey, uri)
+		pool, routeAdded = r.insertRouteKey(routekey, uri)
 		r.RLock()
 	}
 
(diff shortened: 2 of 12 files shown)


maxmoehl:
We've validated this in one of our environments and the metrics now show up with the "refreshed" tag. As expected, the number of route messages labeled as "updated" went down a lot as those are now "refreshed". The logs also behave as expected with the refresh logs (which have a high volume) being not emitted if you set the log level to info.

We also ran the routing-release test suite for gorouter and it passes.

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -88,29 +88,32 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
...
 		}
+	case route.REFRESHED:
+		if r.logger.Enabled(context.Background(), slog.LevelDebug) {
+			r.logger.Debug("endpoint-refreshed", buildSlogAttrs(uri, endpoint)...)
+		}
 	}
-
 }
 
 func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.PoolPutResult {
```
hoffmaen: Can we also introduce a metric for new and removed routes?

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -88,29 +88,32 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
...
+	r.reporter.CaptureRegistryMessage(endpoint, poolPutResult.String())
 
-	if endpointAdded == route.ADDED && !endpoint.UpdatedAt.IsZero() {
+	if poolPutResult == route.ADDED && !endpoint.UpdatedAt.IsZero() {
 		r.reporter.CaptureRouteRegistrationLatency(time.Since(endpoint.UpdatedAt))
 	}
 
-	switch endpointAdded {
+	switch poolPutResult {
 	case route.ADDED:
```
hoffmaen: Can we please have the consts prefixed with `ENDPOINT_`or something similar?
--> `ENDPOINT_ADDED`, `ENDPOINT_UNMODIFIED` etc. 
Otherwise, `route.ADDED` is misleading.

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -169,43 +172,57 @@ func (r *RouteRegistry) Unregister(uri route.Uri, endpoint *route.Endpoint) {
...
+
+	if routeRemoved {
+		r.logger.Info("route-unregistered", slog.Any("uri", uri))
+	} else {
+		r.logger.Info("route-not-unregistered", slog.Any("uri", uri))
+	}
 }
 
-func (r *RouteRegistry) unregister(uri route.Uri, endpoint *route.Endpoint) {
+func (r *RouteRegistry) unregister(uri route.Uri, endpoint *route.Endpoint) (endpointRemoved, routeRemoved bool) {
```
hoffmaen: Can we pleadse have proper types for `endpointRemoved` and `routeRemoved`, similar to the `poolPutResult`?

hoffmaen:
One more thing: We log the registration of new routes very deep in the code (in `insertRouteKey`), but we log the removal of routes in the public function (`Unregister`). Can we have it both on the same level, ideally in the public function?

src/code.cloudfoundry.org/gorouter/registry/registry.go
maxmoehl: Well these are just bools... I'd rather not introduce aliases for `true` and `false` 😄

hoffmaen: Why not, still better than having 
```
return true, false
return false, false
return true, true
```
in the code, where you need to read function signatures to understand the meaning.

peanball: image
This looks pretty clear to me, at least in IntelliJ/GoLand

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -277,35 +280,58 @@ func (p *EndpointPool) Put(endpoint *Endpoint) PoolPutResult {
...
 
-			oldEndpoint := e.endpoint
-			e.endpoint = endpoint
+		return EndpointRefreshed
 
-			if oldEndpoint.PrivateInstanceId != endpoint.PrivateInstanceId {
-				delete(p.index, oldEndpoint.PrivateInstanceId)
-				p.index[endpoint.PrivateInstanceId] = e
-			}
+	case found && !e.endpoint.ModificationTag.SucceededBy(&endpoint.ModificationTag):
```
b1tamara, line 299: Very minor note: I always need time to understand the second condition. Maybe you should move it to a separate function (I know that it will be a one-liner) with a good name like checkIfNoModifications.

src/code.cloudfoundry.org/gorouter/registry/registry.go
b1tamara: I agree with @hoffmaen. The coding would be more readable. But your call.

hoffmaen (approved):

src/code.cloudfoundry.org/gorouter/route/pool.go
maxmoehl: I agree that the condition is not nice, I'd prefer to change it around a bit to make this more readable but there are some edge cases that could change then so I'd rather not touch this. The idea is that the comment in the line below covers the intention of this case. If you need more details than that you will have to look into the function and how it works anyway.

src/code.cloudfoundry.org/gorouter/metrics/metricsreporter_test.go
```diff
@@ -524,6 +524,18 @@ var _ = Describe("MetricsReporter", func() {
 		Expect(count).To(Equal(uint64(5)))
 	})
 
+	It("increments the bad_requests metric", func() {
```
b1tamara: Please adapt the description of the test

src/code.cloudfoundry.org/gorouter/metrics/metricsreporter_test.go
```diff
@@ -524,6 +524,18 @@ var _ = Describe("MetricsReporter", func() {
...
+		metricReporter.CaptureRouteAdded()
+		Expect(batcher.BatchIncrementCounterCallCount()).To(Equal(1))
+		Expect(batcher.BatchIncrementCounterArgsForCall(0)).To(Equal("routes_added"))
+	})
+
+	It("increments the bad_requests metric", func() {
```
b1tamara: Please adapt the description of the test

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -88,32 +88,42 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
 		return
 	}
 
-	endpointAdded := r.register(uri, endpoint)
+	poolPutResult, routeAdded := r.register(uri, endpoint)
```
b1tamara: Maybe routePoolAdded instead of routeAdded? It is clearer then.

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -88,32 +88,42 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
...
+	case route.EndpointAdded:
 		if r.logger.Enabled(context.Background(), slog.LevelInfo) {
 			r.logger.Info("endpoint-registered", buildSlogAttrs(uri, endpoint)...)
 		}
-	case route.UPDATED:
-		if r.logger.Enabled(context.Background(), slog.LevelDebug) {
-			r.logger.Debug("endpoint-registered", buildSlogAttrs(uri, endpoint)...)
+	case route.EndpointUpdated:
+		if r.logger.Enabled(context.Background(), slog.LevelInfo) {
+			r.logger.Info("endpoint-registered", buildSlogAttrs(uri, endpoint)...)
```
b1tamara, line 113: Do we want to update the log message as well? To distinguish between updated and registered?

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -124,88 +134,104 @@ func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.
...
+
+	if endpointRemoved {
+		r.logger.Info("endpoint-unregistered", buildSlogAttrs(uri, endpoint)...)
+	} else {
+		r.logger.Info("endpoint-not-unregistered", buildSlogAttrs(uri, endpoint)...)
+	}
+
+	if routeRemoved {
+		r.logger.Info("route-unregistered", slog.Any("uri", uri))
+		r.reporter.CaptureRouteDeleted()
```
b1tamara: You will miss this metric if logger is not enabled

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -124,88 +134,104 @@ func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.
...
 	r.reporter.CaptureUnregistryMessage(endpoint)
 
+	// WARNING: this only works because there are only info log statements after this point. If you
+	// extend this function, this probably needs to be refactored.
+	if !r.logger.Enabled(context.Background(), slog.LevelInfo) {
+		return
+	}
+
+	if endpointRemoved {
+		r.logger.Info("endpoint-unregistered", buildSlogAttrs(uri, endpoint)...)
```
b1tamara: Maybe you can move logger statements after all if conditions. In the if-condition you can assign what message should be logged like:
```
if endpointRemoved {
    logMsg = "endpoint-unregistered"
} else{
    logMsg = "endpoint-not-unregistered"
}
if routeRemoved {
    logMsg = "route-unregistered"
    r.reporter.CaptureRouteDeleted()
}
...
if r.logger.Enabled(context.Background(), slog.LevelInfo) {
    r.logger.Info(logMsg, buildSlogAttrs(uri, endpoint)...)
}
```

maxmoehl: We want to log route and endpoint independent of each other so that wouldn't work. I've moved the metric further up.

src/code.cloudfoundry.org/gorouter/registry/registry.go
maxmoehl: We decided to adjust the log messages as little as possible to not make this a breaking change.

b1tamara (approved):
LGTM

PR #481, opened 2025-05-07
Validate gorouter, tcp-router and routing-api config on bootstrap VM

mariash:

Summary
---------------
If gorouter, tcp router or routing-api are misconfigured we want canary VM to fail first. Currently, these jobs start process and don't validate config, so canary VM might succeed and this will cause downtime if config is incorrect.

Route-register is going to start (write to pidfile) only after it loads config successfully.

Backward Compatibility
---------------
Breaking Change? **No**


jobs/gorouter/templates/pre-start.erb
```diff
@@ -90,3 +90,7 @@ ulimit -n <%= p("router.max_open_files") %>
 # Add jq to path
 cp /var/vcap/jobs/gorouter/bin/setup-jq /etc/profile.d/jq.sh
 chown root:vcap /etc/profile.d/jq.sh
+
+<% if spec.bootstrap == true %>
```
ebroberson, line 94: Do we want to do this for ONLY the canary?  Is there any downside to making it run on every instance?  What happens if they don't use a canary?

mariash: Bosh should always have 1 instance that is bootstrap instance and in case of canary it will be the bootstrap one. If there is no canary then it will just run it on one of them.

ebroberson: Awesome, thanks for the clarification!

ebroberson (approved):
LGTM

PR #482, opened 2025-05-12
fix: properly order ginkgo arguments

maxmoehl:

Summary
---------------
Packages needs to be specified last, with 60d74d0 we introduced the `--nodes` argument as the last one which works as long as you don't try to specify a sub-package.

This commit re-orders the arguments so that the package is last argument again.

Backward Compatibility
---------------
Breaking Change? **No**


plowin (approved):
thx!

PR #483, opened 2025-05-15
Revert 100-continue specific proxy now that it's handled correctly in golang

geofffranks:

Summary
---------------
Reverts the changes made to work around a 100-continue bug that golang has since patched.

See https://github.com/cloudfoundry/routing-release/issues/470.

Backward Compatibility
---------------
Breaking Change? no


peanball:
In your commit you note that the integration tests are kept. Should / can we also keep the tests in proxy_test.go?

geofffranks:
> In your commit you note that the integration tests are kept. Should / can we also keep the tests in proxy_test.go?

Those were tied to making sure the  proxy.go was doing the right things when `expect100ContinueRProxy` was being used as the proxy. Since that's gone, its tests are too. The tests would have been a duplicate of the original `rproxy`'s tests though so it still has its own tests covering the same things.

maxmoehl (approved):
Thank you for cleaning this up!

Test suite looks good but the package syncing produces a diff. Up to you if you want to include it in this PR or leave it to the CI job to fix that. I didn't manage to deploy it to our env.

PR #492, opened 2025-06-02
Remove old property: router.max_header_kb

kart2bc:

Summary
---------------
Remove old property: router.max_header_kb

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -287,20 +287,13 @@ properties:
   router.route_services_strict_signature_validation:
       description: "Enforce strict validation of a route service signature"
       default: false
-  router.max_header_kb:
-    description: |
-        Deprecated, use router.max_request_header_kb instead which is equivalent to this option.
-    default: 1024 # 1Mb
   router.max_request_header_kb:
     description: |
         This value controls the maximum number of bytes (in KB) the gorouter will read
         parsing the request header's keys and values, including the request
         line. It does not limit the size of the request body. Requests with
         larger headers will result in a 431 status code. Must be between 1 and 1024kb.
-        Note: This value takes precedence over router.max_header_kb but has no default to not break
-        existing setups. If you previously configured router.max_header_kb it is recommended to
-        switch to this property instead.
-    example: 1024 # 1Mb
+    default: 1024 # 1Mb
   router.max_response_header_kb:
     description: |
         This value controls the maximum number of bytes (in KB) the gorouter will read
--- a/jobs/gorouter/templates/gorouter.yml.erb
+++ b/jobs/gorouter/templates/gorouter.yml.erb
@@ -76,14 +76,9 @@ def status_tls
 end
 
 def max_request_header_bytes
-  # See property description for details.
-  if_p("router.max_request_header_kb") do |kb|
-    return kb * 1024
-  end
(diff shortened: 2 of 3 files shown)


spec/gorouter_templates_spec.rb
```diff
@@ -344,25 +343,24 @@
...
-            deployment_manifest_fragment['router']['max_header_kb'] = 10
+            deployment_manifest_fragment['router']['max_request_header_kb'] = 10
           end
           it 'should set max_request_header_bytes according to it' do
             expect(parsed_yaml['max_request_header_bytes']).to eq(10_240)
           end
         end
 
-        context 'when max_header_kb and max_request_header_kb are set' do
+        context 'when max_request_header_kb and max_request_header_kb are set' do
```
ameowlia: you can delete this test since it doesn't make sense in this new world without max_header_kb

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -80,13 +80,6 @@ def max_request_header_bytes
   if_p("router.max_request_header_kb") do |kb|
```
ameowlia: Now that the max_request_header_kb has a default value (nice job changing that, it was the correct thing to do), we no longer need to use `if_p` here. `if_p` is only if there is a chance the property might not be set. Instead you can always grab the property with `kb = p("router.max_request_header_kb")`

jobs/gorouter/spec
```diff
@@ -287,20 +287,13 @@ properties:
...
-  router.max_header_kb:
-    description: |
-        Deprecated, use router.max_request_header_kb instead which is equivalent to this option.
-    default: 1024 # 1Mb
   router.max_request_header_kb:
     description: |
         This value controls the maximum number of bytes (in KB) the gorouter will read
         parsing the request header's keys and values, including the request
         line. It does not limit the size of the request body. Requests with
         larger headers will result in a 431 status code. Must be between 1 and 1024kb.
```
ameowlia, line 295: This says: Must be between 1 and 1024kb.

Can you add tests + logic to enforce this?

ameowlia (changes requested):

jobs/gorouter/spec
kart2bc: I have restored the [condition check](https://github.com/cloudfoundry/routing-release/pull/492/files#diff-c0898fc9a127ca8eb4ed3b1677c109b3fc96ccd073a59218a230dce5d6ff7483L85) . So it will throw an error if the value is not within the limit. 
So it may not require a test case for the values  x < 1 and x > 1024. Other test cases are already in place.

ameowlia: I don't see tests cases for when it is too high or low. Can you link to them?

kart2bc: We will not encounter a value being too low or high as it will throw an error.  
`raise "Invalid router.max_request_header_kb value. Must be between 1 and 1,024 kb"`
Perhaps, are there any other ways, the user can bypass this check?

ameowlia (approved):

maxmoehl:
Thank you! Was still on my list but never got around to it 🙈

PR #501, opened 2025-07-16
DRAFT: upgrade counterfeiter

neowulf:
add counterfeiter:generate directive

* add Makefile for repeatable go:generate commands
* upgrade counterfeiter
* add counterfeiter by using `go tool counterfeiter`
* clean up inconsistencies in how the files were generated

Depends on https://github.com/cloudfoundry/routing-api/pull/74/files

Summary
---------------

Backward Compatibility
---------------
Breaking Change? **No**


src/code.cloudfoundry.org/Makefile
```diff
@@ -0,0 +1,8 @@
+.PHONY: clean
+clean:
+	find . -type d -name fakes -exec rm -rf {} \;  || true
+
+.PHONY: generate
+generate:
+	go generate ./...
+	gsed -i '/new(FakeVarzReporter)/i \/\/lint:ignore SA1019 - auto-generated fake will go away when Varz goes away' gorouter/metrics/fakes/fake_varzreporter.go
```
maxmoehl, line 8: Adding a make file just for this one use-case seems a bit odd.`go generate ./...` probably doesn't need one and could we replace the injected linter ignore directive with a staticcheck config that ignores the check for this particular (or all generated) file(s)?

https://staticcheck.dev/docs/configuration/

src/code.cloudfoundry.org/gorouter/accesslog/file_and_loggregator_access_logger.go
```diff
@@ -12,7 +12,7 @@ import (
 	log "code.cloudfoundry.org/gorouter/logger"
 )
 
-//go:generate counterfeiter -o fakes/accesslogger.go . AccessLogger
+//counterfeiter:generate -o fakes/accesslogger.go . AccessLogger
```
maxmoehl, line 15: I find it a bit confusing that some need both directives and others only one but I get that we only need to run this once per package. Should we maybe move the `//go:generate` to a `gen.go` file then?

PR #502, opened 2025-07-22
Added property to accept certificates for tls termination at tcp router

AshishNaware:

Summary
---------------

This PR introduces new property called frontend_tls to accept certs and keys that are used to terminate tls traffic at tcp router.

Backward Compatibility
---------------
Breaking Change? **Yes/No**

No


jobs/tcp_router/templates/tcp_router.yml.erb
```diff
@@ -91,3 +91,50 @@ backend_tls:
...
+
+<%
+  frontend_tls = p('tcp_router.frontend_tls', [])
+
+  if frontend_tls&.any?
+    frontend_tls.each_with_index do |cert_pair, i|
+      unless cert_pair.is_a?(Hash) &&
+             cert_pair.key?('cert_chain') &&
+             cert_pair.key?('private_key')
+        raise "frontend_tls[#{i}] must include both cert_chain and private_key"
```
geofffranks: The bosh team might be wanting us to get away from adding validations like this during template rendering, and prefer us doing it in a pre-start script. We've done something like this for a couple other components (i think gorouter in routing-release is one). I asked to see if that's still their desire.

But if there are other templates in this release that do things the old way in ERBs, I guess this is fine to keep like that, and we'll fix it all when we get to fixing tcp-router?

jobs/tcp_router/templates/tcp_router.yml.erb
```diff
@@ -91,3 +91,50 @@ backend_tls:
...
+end
+-%>
+
+frontend_tls:
+<% p('tcp_router.frontend_tls', []).each do |pair| -%>
+  - cert_chain: |
+<%= indent(pair['cert_chain'], 6) %>
+    private_key: |
+<%= indent(pair['private_key'], 6) %>
+<% end -%>
```
geofffranks, line 106: Why are we passing the tls configs as tcp-router config, and also in the pem file listed below? Shouldn't we only need them in one of those two places?

jobs/tcp_router/templates/tcp_router_frontend_client_cert_and_key.pem.erb
```diff
@@ -0,0 +1,8 @@
+<%
+frontend_tls = p('tcp_router.frontend_tls', [])
+frontend_tls.each_with_index do |tls, i|
+%>
+<%= tls['cert_chain'] %>
+<%= tls['private_key'] %>
+
```
geofffranks: Have you validated that haproxy is smart enough to be able to associate each private key with its chain certs when specified this way? 

Especially if we have a 2 or 3-cert chaing associated with a single key?

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -81,6 +86,21 @@ func (c *Config) initConfigFromFile(path string) error {
...
+		}
+
+	} else {
+		c.FrontendTLS = append(c.FrontendTLS, FrontendTLSConfig{
+			CertChain:  "",
+			PrivateKey: "",
+		})
+	}
+
 	if c.BackendTLS.Enabled {
```
geofffranks, line 151: I was expecting more tcp-router logic here (to do stuff with the new values), but 🤞 thats coming in a different commit/pr?

Can we get tests to ensure the config validation is working correctly when certs are provided, when frontend-tls is provided with no certs and when no frontend-tls is provided at all?

jobs/tcp_router/templates/tcp_router.yml.erb
geofffranks: Yeah, it sounds like they are wanting things to be validated at pre-start time.

Here's how gorouter is doing it:

https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/templates/pre-start.erb#L94-L96

https://github.com/cloudfoundry/routing-release/blob/develop/src/code.cloudfoundry.org/gorouter/cmd/config-validator/main.go

jobs/tcp_router/templates/tcp_router_frontend_client_cert_and_key.pem.erb
AshishNaware: ~~Hi @geofffranks - Thanks for the feedback. I was discussing this with Siva earlier today. I will push changes that will create separate directories with corresponding cert chain and private key. We also discussed that instead of passing the actual cert and key, it should pass the path to that directory. Could you help me point out where is the script/ or where do I add the script that persists those certs and keys in `config/keys/tcp-router/frontend/` directory?~~

I have pushed new changes regarding this

jobs/tcp_router/templates/tcp_router.yml.erb
AshishNaware: I moved validations to pre-start in latest commit

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -156,3 +223,20 @@ func (c *Config) initConfigFromFile(path string) error {
 
 	return nil
 }
+
+func certHasSAN(cert *x509.Certificate) bool {
+	hasSANExtension := false
+	for _, ext := range cert.Extensions {
+		if ext.Id.String() == "2.5.29.17" {
```
neowulf, line 230: Some documentation here may help.

Where did this id `2.5.29.17` come from? Is this a well known constant in the go library perhaps? What would happen if this value changes? How do we update this value?

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -60,6 +73,13 @@ func New(path string) (*Config, error) {
 	return c, nil
 }
 
+func (c *Config) FrontendTLSJobBasePath() string {
+	if bp := os.Getenv("FRONTEND_TLS_BASE_PATH"); bp != "" {
+		return bp
+	}
+	return "/var/vcap/jobs/tcp_router/config/keys/tcp-router/frontend"
```
neowulf, line 80: do we have to worry about proper permissions? Copying the sibling's (`backend`) permissions:

```
chown -R root:vcap /var/vcap/jobs/tcp_router/config/certs/tcp-router/frontend
chmod -R 640 	  /var/vcap/jobs/tcp_router/config/certs/tcp-router/frontend
chmod      750 	  /var/vcap/jobs/tcp_router/config/certs/tcp-router/frontend
```

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -81,6 +101,53 @@ func (c *Config) initConfigFromFile(path string) error {
...
+			hasSAN := certHasSAN(cert)
+			if !hasSAN {
+				return fmt.Errorf("frontend_tls[%d].cert_chain must include a subjectAltName extension", i)
+			}
+
+			dirPath := filepath.Join(basePath, name)
+			os.MkdirAll(dirPath, 0755)
+
+			certFilePath := filepath.Join(dirPath, fmt.Sprintf("%s.cert.pem", name))
+			keyFilePath := filepath.Join(dirPath, fmt.Sprintf("%s.key.pem", name))
```
neowulf:
```suggestion
			certFilePath := filepath.Join(dirPath, fmt.Sprintf("%s.pem", name))
			keyFilePath := filepath.Join(dirPath, fmt.Sprintf("%s.pem.key", name))
```

The key is inferred by haproxy by suffixing `.key` to the cert's file name

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -156,3 +223,20 @@ func (c *Config) initConfigFromFile(path string) error {
...
+			break
+		}
+	}
+
+	hasSANEntries := len(cert.DNSNames) > 0 ||
+		len(cert.EmailAddresses) > 0 ||
+		len(cert.IPAddresses) > 0 ||
+		len(cert.URIs) > 0
+
+	return hasSANExtension || hasSANEntries
```
neowulf: If DNSNames are missing, how will the traffic get routed using a DNS name?

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
blancobrenden: This is the OID of the SAN field coming from the ASN1 spec. It will never change. https://oidref.com/2.5.29.17

I _think_ that the check that computes `hasSANEntries` below is sufficient / redundant, but not 100% sure.

AshishNaware: I added this check to be consistent with the gorouter https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/templates/gorouter.yml.erb#L445

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
AshishNaware: modified this check to verify if the dns entries are present

AshishNaware:
closing this PR as these changes will be included in this consolidated PR - https://github.com/cloudfoundry/routing-release/pull/503

PR #504, opened 2025-07-29
Log redundant route/endpoint unregistration on Debug level

hoffmaen:

Summary
When unregistering an endpoint, e.g. when scaling down the number of instance, simply rescheduling an application on another cell, or when app instances are stopped, crashed, or removed, route-emitter repeatedly sends unregistration messages via NATS towards gorouter. The number of repeats for sending unregistration messages is [configurable](https://github.com/cloudfoundry/route-emitter/blob/26f4ea327c50ff95828162de3f6db8f14f25f49c/cmd/route-emitter/config/config.go#L64). This ensures reliable route cleanup even if Gorouter misses earlier messages. It also prevents stale routes 
from remaining registered.

Gorouter logs each of these messages on `Info` level, even the redundant messages, when the endpoint and/or route has already been unregistered. This leads to a huge amount of redundant log messages. This change proposes to keep the logging level at `Info` for the actual unregistration, but change it to `Debug` for subsequent redundant messages.  

Backward Compatibility
---------------
Breaking Change? **Questionable**
While changing the log level may be considered a breaking change, the affected log messages are redundant and do not provide any value.

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/registry/registry_test.go
+++ b/src/code.cloudfoundry.org/gorouter/registry/registry_test.go
@@ -845,6 +845,14 @@ var _ = Describe("RouteRegistry", func() {
 				Eventually(logger).Should(gbytes.Say(`"log_level":1.*route-unregistered.*a\.route`))
 			})
 
+			It("logs the redundant route and endpoint unregistration at debug level", func() {
+				r.Unregister("a.route", fooEndpoint)
+				Eventually(logger).Should(gbytes.Say(`"log_level":1.*endpoint-unregistered.*a\.route.*192\.168\.1\.1`))
+				Eventually(logger).Should(gbytes.Say(`"log_level":1.*route-unregistered.*a\.route`))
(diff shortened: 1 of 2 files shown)


src/code.cloudfoundry.org/gorouter/registry/registry_test.go
```diff
@@ -845,6 +845,14 @@ var _ = Describe("RouteRegistry", func() {
 				Eventually(logger).Should(gbytes.Say(`"log_level":1.*route-unregistered.*a\.route`))
 			})
 
+			FIt("logs the redundant route and endpoint unregistration at debug level", func() {
```
maxmoehl: Please remove the F here.

maxmoehl (changes requested):

maxmoehl:
I would also argue that this is not a breaking change. IMO breaking changes in logging are more related to changing existing fields / messages which can trip existing log parsing rules which (hopefully) shouldn't be the case here.

maxmoehl (approved):

PR #508, opened 2025-08-28
tcp_router: fix cert directory permissions

neowulf:
- [ x ] Read the [Contributing document](../blob/-/.github/CONTRIBUTING.md).

Summary
---------------
- ensure the tcp-router configuration also has the derived frontend_tls_config
- ensure the directory permissions are root:vcap

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/jobs/tcp_router/templates/pre-start.erb
+++ b/jobs/tcp_router/templates/pre-start.erb
@@ -1,5 +1,4 @@
 #!/bin/bash -e
 
-<% if spec.bootstrap == true %>
+# certs need to be written to the filesystem on all VMs
 /var/vcap/packages/tcp_router/bin/config-validator -config /var/vcap/jobs/tcp_router/config/tcp_router.yml -enable-cert-creation true
-<% end %>
--- a/jobs/tcp_router/templates/pre-start.erb
+++ b/jobs/tcp_router/templates/pre-start.erb
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 
-# certs need to be written to the filesystem on all VMs
+# This pre-start script runs (1) config validation and (2) creates frontend certs on the VM (if any)
 /var/vcap/packages/tcp_router/bin/config-validator -config /var/vcap/jobs/tcp_router/config/tcp_router.yml -enable-cert-creation true
--- a/src/code.cloudfoundry.org/cf-tcp-router/config/config.go
+++ b/src/code.cloudfoundry.org/cf-tcp-router/config/config.go
@@ -102,6 +102,12 @@ func resolveGroupID(primary, fallback string) (int, error) {
 	return gid, nil
 }
 
+// initConfigFromFile initializes the config and write the certs to the filesystem when enableCertCreation is true.
+//
+// enableCertCreation is set to true during pre-start on all VMs. During pre-start initConfigFromFile is run
+// to (1) validate the config provided and (2) create the FrontendTLSJob certs on the filesystem if provided.
+//
+// enableCertCreation is set to false when bosh starts the tcp-router job.
 func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error {
 	b, err := os.ReadFile(path)
 	if err != nil {
@@ -121,70 +127,56 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error 
 	}
 
 	if enableCertCreation && len(c.FrontendTLSJob) > 0 {
-		owner, err := user.Lookup("root")
-		if err != nil {
-			return err
+		// remove existing directory which will clean out old certs if necessary
+		if err := os.RemoveAll(c.FrontendTLSJobBasePath()); err != nil && !os.IsNotExist(err) {
+			return fmt.Errorf("directory could not be removed: %w", err)
 		}
 
-		uid, err := strconv.Atoi(owner.Uid)
-		if err != nil {
-			return err
+		if err := c.createDir(c.FrontendTLSJobBasePath()); err != nil {
+			return fmt.Errorf("could not create directory: %w", err)
+		}
+	}
+
+	for i, cert := range c.FrontendTLSJob {
+		// validations
+		name := strings.TrimSpace(cert.Name)
+		if name == "" {
+			return fmt.Errorf("frontend_tls[%d]: empty name", i)
 		}
 
-		gid, err := resolveGroupID("vcap", "root")
-		if err != nil {
-			return err
-		}
-		var outputs []FrontendTLSConfig
-		basePath := c.FrontendTLSJobBasePath()
-		for i, cert := range c.FrontendTLSJob {
-
-			name := strings.TrimSpace(cert.Name)
-			certChain := strings.TrimSpace(cert.CertChain)
-			privateKey := strings.TrimSpace(cert.PrivateKey)
-
-			if name == "" || certChain == "" || privateKey == "" {
-				return fmt.Errorf("frontend_tls[%d] must include name, cert_chain, and private_key", i)
-			}
-
-			block, _ := pem.Decode([]byte(certChain))
-			if block == nil {
-				return errors.New("failed to parse PEM block")
-			}
-			cert, err := x509.ParseCertificate(block.Bytes)
-			if err != nil {
-				return err
-			}
-
-			hasSAN := certHasSAN(cert)
-			if !hasSAN {
-				return fmt.Errorf("frontend_tls[%d].cert_chain must include a subjectAltName extension", i)
-			}
-
-			dirPath := filepath.Join(basePath, name)
-			if err := os.MkdirAll(dirPath, 0750); err != nil {
-				return err
-			}
-
-			certFilePath := filepath.Join(dirPath, fmt.Sprintf("%s.pem", name))
-			keyFilePath := filepath.Join(dirPath, fmt.Sprintf("%s.pem.key", name))
-
-			if err := writeFile(certFilePath, []byte(certChain), 0750, uid, gid); err != nil {
-				return err
-			}
-
-			if err := writeFile(keyFilePath, []byte(privateKey), 0750, uid, gid); err != nil {
-				return err
-			}
-
-			outputs = append(outputs, FrontendTLSConfig{
-				Enabled:        true,
-				CertificateDir: dirPath,
-			})
+		certChain := strings.TrimSpace(cert.CertChain)
+		if certChain == "" {
+			return fmt.Errorf("frontend_tls[%d]: empty cert_chain", i)
 		}
 
-		c.FrontendTLS = outputs
+		privateKey := strings.TrimSpace(cert.PrivateKey)
+		if privateKey == "" {
+			return fmt.Errorf("frontend_tls[%d]: empty private_key", i)
+		}
 
+		// check if san is present and that the cert chain is valid
+		if err := certHasSAN(cert.CertChain); err != nil {
+			return fmt.Errorf("frontend_tls[%d]: %w", i, err)
+		}
+
+		dirPath := filepath.Join(c.FrontendTLSJobBasePath(), name)
+
+		// write certs to the disk
+		if enableCertCreation {
+			if err := c.createDir(dirPath); err != nil {
+				return fmt.Errorf("frontend_tls[%d]: %w", i, err)
+			}
+
+			if err := c.writeCertsToDisk(dirPath, name, cert, privateKey); err != nil {
+				return fmt.Errorf("frontend_tls[%d]: %w", i, err)
+			}
+		}
+
+		// update the config
+		c.FrontendTLS = append(c.FrontendTLS, FrontendTLSConfig{
+			Enabled:        true,
+			CertificateDir: dirPath,
+		})
 	}
 
 	if c.BackendTLS.Enabled {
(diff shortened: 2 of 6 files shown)


src/code.cloudfoundry.org/cf-tcp-router/config/config_test.go
```diff
@@ -323,6 +323,41 @@ var _ = Describe("Config", Serial, func() {
...
+				Expect(cfg.FrontendTLS[0]).To(Equal(config.FrontendTLSConfig{
+					Enabled:        true,
+					CertificateDir: filepath.Join(tmpDir, "prod"),
+				}))
+
+				Expect(cfg.FrontendTLS[1]).To(Equal(config.FrontendTLSConfig{
+					Enabled:        true,
+					CertificateDir: filepath.Join(tmpDir, "dev"),
+				}))
+			})
```
ameowlia: Why are certs being added when enableCertCreation is False?

src/code.cloudfoundry.org/cf-tcp-router/config/config_test.go
```diff
@@ -272,7 +272,7 @@ var _ = Describe("Config", Serial, func() {
 			})
 		})
 
-		Context("with invalid cert and key", func() {
+		Context("with invalid cert and key and enableCertCreation set to True", func() {
```
ameowlia: This context SAYS it is about when there is an invalid cert and key, but a few lies down it says it is valid and everything works.

ameowlia: I think you are missing a whole host of tests for when the cert/keys are invalid.

Every error case in your code should have its own unit test.

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -120,63 +125,43 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
-		uid, err := strconv.Atoi(owner.Uid)
-		if err != nil {
-			return err
-		}
-
-		gid, err := resolveGroupID("vcap", "root")
-		if err != nil {
-			return err
-		}
+	if len(c.FrontendTLSJob) > 0 {
```
ameowlia: Shouldn't there be an overarching if for `f enableCertCreation` ? And there should be an error case if `len(c.FrontendTLSJob) == 0`, which I assume is an invalid state that would not be currently caught.

ameowlia (changes requested):
After you address the review comments, can you also provide an overview of which cases you tested manually by deploying to a real CF with the output?

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
neowulf: these are two different concerns and are executed in two different contexts. 

One context is when the bosh job is running which will create the directories and write the certs. This is when [`enableCertCreation` is true](https://github.com/cloudfoundry/routing-release/blob/401e03290261d0c5bd387cb04a3cba71f7b1a401/jobs/tcp_router/templates/pre-start.erb#L4). 

Second context is when the monit job starts up which updates the config. This is when `enableCertCreation` is false.

neowulf: if there are no `FrontendTLSJob`s - that means this feature is disabled

src/code.cloudfoundry.org/cf-tcp-router/config/config_test.go
neowulf: Please see comment:
https://github.com/cloudfoundry/routing-release/pull/508/files#r2316441969

src/code.cloudfoundry.org/cf-tcp-router/config/config_test.go
neowulf: true, let me add some more coverage for the error cases as well

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -102,6 +103,10 @@ func resolveGroupID(primary, fallback string) (int, error) {
 	return gid, nil
 }
 
+// initConfigFromFile initializes the config and write the certs to the filesystem when enableCertCreation is true.
+//
+//	enableCertCreation is usually set to true in the tcp_router pre-start job which will create the certs on the filesystem.
```
ameowlia:
```suggestion
//	enableCertCreation is set to true during pre-start on the canary VM. During pre-start initConfigFromFile is run to (1) validate the config provided and (2) create the FrontendTLSJob certs on the filesystem if provided.
```

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -102,6 +103,10 @@ func resolveGroupID(primary, fallback string) (int, error) {
...
+//	enableCertCreation is set to false which is the normal mode to run the tcp_router which will not modify the filesystem.
```
ameowlia:
```suggestion
//	enableCertCreation is set to false when bosh starts the tcp-router job.
```

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
ameowlia: Based on [this code](https://github.com/cloudfoundry/routing-release/blob/401e03290261d0c5bd387cb04a3cba71f7b1a401/jobs/tcp_router/templates/pre-start.erb#L4). Certs will only be added to the file system on the canary VM.

Is that intended?

neowulf: Spoke offline with @ameowlia - certs must be added to all VMs. I will make that change.

neowulf: Regarding the different contexts - if we attempt to write the certs during the normal bootup, the following error causes the job to fail:

```
==> /var/vcap/sys/log/tcp_router/tcp_router.stdout.log <==
Wed Sep  3 04:09:33 AM UTC 2025: Starting HAProxy
Wed Sep  3 04:09:33 AM UTC 2025: Successfully started HAProxy
uaa.service.cf.internal has address 88.0.0.31
{"timestamp":"2025-09-03T04:09:33.466539294Z","level":"info","source":"tcp-router","message":"tcp-router.starting","data":{}}
{"timestamp":"2025-09-03T04:09:33.476675165Z","level":"error","source":"tcp-router","message":"tcp-router.failed-to-unmarshal-config-file","data":{"error":"frontend_tls[0]: Error changing ownership: chown /var/vcap/jobs/tcp_router/config/certs/tcp-router/frontend: read-only file system"}}
```

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
neowulf: I have slightly modified the suggestion to indicate we want the certs to be created on all vms

neowulf:
Tested the following scenarios on live cf for the frontend TLS config :
1. no certs
2. 1 cert
3. multiple certs
4. multiple instances

jobs/tcp_router/templates/pre-start.erb
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash -e
 
-<% if spec.bootstrap == true %>
+# certs need to be written to the filesystem on all VMs
```
ameowlia: I regret saying that we could get rid of this if and have decided I don't like this change. I think this change overloads the config-validator. 

There IS a reason for config validation to only run on 1 VM: the config is the same on every VM. Only 1 VM needs to check the config and fail if something is wrong. There is NOT a reason for cert creation to be run during pre-start. 

Please change the PR so that the certs are created at runtime and not during pre-start.

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -120,71 +127,42 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
-		owner, err := user.Lookup("root")
-		if err != nil {
-			return err
+	for i, cert := range c.FrontendTLSJob {
+		// validations
+		name := strings.TrimSpace(cert.Name)
+		certChain := strings.TrimSpace(cert.CertChain)
+		privateKey := strings.TrimSpace(cert.PrivateKey)
+		if name == "" || privateKey == "" || certChain == "" {
+			return fmt.Errorf("frontend_tls[%d]: invalid name, cert_chain, or private_key", i)
```
ameowlia: This error message should be improved:

1. the error says "invalid", but all this actually tests for is if the items are missing.
2. the error does not specify WHICH item is the issue. Please create a different error message for each item.

ameowlia: 👍 However, just like to note that I do see tests for all these of these error cases.

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -262,28 +240,97 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
 		}
 	}
 
-	hasDNSEntries := len(cert.DNSNames) > 0
+	if len(cert.DNSNames) > 0 {
+		return nil
+	}
 
-	return hasSANExtension || hasDNSEntries
+	return fmt.Errorf("cert_chain must include either a subjectAltName extension or DNSNames")
```
ameowlia, line 330: 👍 tested

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -262,28 +240,97 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
-			hasSANExtension = true
-			break
+			return nil
 		}
 	}
 
-	hasDNSEntries := len(cert.DNSNames) > 0
+	if len(cert.DNSNames) > 0 {
+		return nil
+	}
```
ameowlia, line 328: I don't see a test for this case.

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -262,28 +240,97 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
+
+func certHasSAN(certChain string) error {
+	if certChain == "" {
+		return errors.New("cert chain is empty")
+	}
+
+	certChain = strings.TrimSpace(certChain)
+	block, _ := pem.Decode([]byte(certChain))
+	if block == nil {
+		return errors.New("failed to parse PEM block")
```
ameowlia, line 318: 👍 tested

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -262,28 +240,97 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
+	if err != nil {
+		return 0, 0, err
+	}
+
+	return uid, gid, nil
+}
+
+func certHasSAN(certChain string) error {
+	if certChain == "" {
+		return errors.New("cert chain is empty")
```
ameowlia: not tested

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -262,28 +240,97 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
+		return errors.New("failed to parse PEM block")
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return err
+	}
+
 	for _, ext := range cert.Extensions {
 		if ext.Id.String() == "2.5.29.17" {
```
ameowlia: Add a comment about what "2.5.29.17" is and why you are checking for it.

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -262,28 +240,97 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
+
+	certChain = strings.TrimSpace(certChain)
+	block, _ := pem.Decode([]byte(certChain))
+	if block == nil {
+		return errors.New("failed to parse PEM block")
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return err
```
ameowlia: Not tested

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
```diff
@@ -262,28 +240,97 @@ func (c *Config) initConfigFromFile(path string, enableCertCreation bool) error
...
+	}
+
+	if err := writeFile(keyFilePath, []byte(privateKey), 0750, uid, gid); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Config) getUIDGID() (int, int, error) {
```
ameowlia, line 296: is there anyway to tests this function?

src/code.cloudfoundry.org/cf-tcp-router/config/config_test.go
```diff
@@ -296,51 +304,47 @@ var _ = Describe("Config", Serial, func() {
 				}))
 			})
 
-			It("writes the correct cert and key files with correct permissions", func() {
-				for i, name := range []string{"prod", "dev"} {
+			It("does not write the correct cert and key files as enableCertCreation is false", func() {
```
ameowlia:
```suggestion
			It("does not write the cert and key files as enableCertCreation is false", func() {
```

ameowlia: I am confused about the wording of the test above (but I can't comment on it because it didn't change in this PR). 

The test on line 292 says:
`It("adds the certs and keys to the expected directories", func() {`

Then this test says `does not write the correct cert and key files as enableCertCreation is false"`
Those seem contradictory to me. 

Please clarify.

ameowlia (changes requested):

jobs/tcp_router/templates/pre-start.erb
neowulf: it's not possible to create the certs as this error is thrown out:
https://github.com/cloudfoundry/routing-release/pull/508#discussion_r2317709119

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
neowulf: This function is looking up the user id and the group id. I am not sure if any part can be mocked out to create a useful test.

Thoughts?

jobs/tcp_router/templates/pre-start.erb
ameowlia: To write during runtime you can do it in the data dir and setup a writeable path via BPM [here](https://github.com/cloudfoundry/routing-release/blob/9956cbcb36fe0c80f404871fbd4782b6655ce414/jobs/tcp_router/templates/bpm.yml.erb#L12-L19). For example "/var/vcap/data/tcp_router/config" is already writable.

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
neowulf: Spoke offline.

> To write during runtime you can do it in the data dir and setup a writeable path via BPM [here](https://github.com/cloudfoundry/routing-release/blob/9956cbcb36fe0c80f404871fbd4782b6655ce414/jobs/tcp_router/templates/bpm.yml.erb#L12-L19). For example "/var/vcap/data/tcp_router/config" is already writable.

Decision was made to create a directory inside `/var/vcap/data/tcp_router/config` which holds the certificates and to be written on bootup

src/code.cloudfoundry.org/cf-tcp-router/config/config_test.go
neowulf: fixed

neowulf: True, I have updated the wording to clarify what was going on:
```
			It("updates the config with the expected directories", func() {
```

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
neowulf: removing this check as it's not needed

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
neowulf: fixed

src/code.cloudfoundry.org/cf-tcp-router/config/config.go
neowulf: fixed

jobs/tcp_router/templates/pre-start.erb
ameowlia:
```suggestion
# This pre-start script runs (1) config validation and (2) creates frontend certs on the VM (if any)
```

ameowlia (approved):

Issue #505, opened 2025-08-01
Implement hash-based routing as a per-route load balancing Algorithm in gorouter


Soha-Albaghdady:
### Proposed Change

As a CF app developer,
I want to leverage hash-based routing in my application scenarios,
So that incoming traffic can be efficiently distributed across various instances based on predefined hash values, ensuring optimal performance and reliability.

In n [RFC-0042](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0042-hash-based-routing.md) "Implementing a Hash-Based Load Balancing Algorithm for CF Routing" we decided to introduce hash-based routing as a load balancing algorithm to be used on a per-route basis. This approach involves the Gorouter using the hash of a request header to make routing decisions. The focus is on distributing users across instances instead of individual requests.

### Acceptance criteria

- Gorouter can parse the new fields containing the per-route options related to hash-routing
- Gorouter can calculate a hash based on the header
- Gorouter can apply the new load-balancing algorithm to make routing decisions for routes that provide fields related to the hash-routing
- Gorouter can assess the current number of in-flight requests across all application instances and ensure the balance factor is not exceeded before routing to an instance
- Gorouter should not incur any performance impact when no application use hash-routing

plowin:
Discussion around performance-testing and feature-flags is taking place via Slack:
https://cloudfoundry.slack.com/archives/C02HNDJB31R/p1757068276784729

The decisions will be documented here.

hoffmaen:
We would like to discuss the following edge cases. I added our proposals for each case.

1. [`X-CF-App-Instance`](https://docs.cloudfoundry.org/concepts/http-routing.html#app-instance-routing) vs Hash-Based Routing: The `X-CF-App-Instance` header's purpose is to route requests to a specific app instance for debugging. Thus, the header should have precedence.
2. [Sticky Sessions (Session Affinity)](https://docs.cloudfoundry.org/concepts/http-routing.html#sessions) vs. Hash-Based Routing: The sticky session cookie should have precedence and the request should be routed to the instance stored in the cookie. This is the case when an application sets the `JSESSIONID` cookie, which is then sent by the client, together with Gorouter's `__VCAP_ID__` cookie. This allows a consistent behavior across load balancing algorithms.
3. Requests do not have the header set that is configured for the hash-based route option. In this case, we should fall back to the global default load balancing algorithm (Round Robin or Least Connection).
[...]

PR #509, opened 2025-09-01
[Hash-Based Load Balancing] Implement registration message types

a18e:
See: https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0042-hash-based-routing.md
See: https://github.com/cloudfoundry/routing-release/issues/505
Co-authored-by: tamara.boehm@sap.com

Summary
---------------

- Completes Task 1 of [Tracking issue](https://github.com/cloudfoundry/routing-release/issues/505): 
"Gorouter can parse the new fields containing the per-route options related to hash-routing"
- Introduces GlobalLoadBalancingAlgorithms to keep allowed global load balancing strategies separate from the load balancing strategies allowed on the application level (via per-route options)

- Reorganize the tests related to the per_route options in the subscriber_test.go

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/config/config.go
+++ b/src/code.cloudfoundry.org/gorouter/config/config.go
@@ -39,7 +39,7 @@ const (
 )
 
 var (
-	GlobalLoadBalancingAlgorithms   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}
+	GlobalLoadBalancingStrategies   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}
 	LoadBalancingStrategies         = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC, LOAD_BALANCE_HB}
 	AZPreferences                   = []string{AZ_PREF_NONE, AZ_PREF_LOCAL}
 	AllowedShardingModes            = []string{SHARD_ALL, SHARD_SEGMENTS, SHARD_SHARED_AND_SEGMENTS}
@@ -598,7 +598,7 @@ func DefaultConfig() (*Config, error) {
 }
 
 func IsGlobalLoadBalancingAlgorithmValid(lbAlgo string) bool {
-	return slices.Contains(GlobalLoadBalancingAlgorithms, lbAlgo)
+	return slices.Contains(GlobalLoadBalancingStrategies, lbAlgo)
 }
 
 func IsLoadBalancingAlgorithmValid(lbAlgo string) bool {
@@ -762,7 +762,7 @@ func (c *Config) Process() error {
 	}
 
 	if !IsGlobalLoadBalancingAlgorithmValid(c.LoadBalance) {
-		return fmt.Errorf("Invalid global load balancing algorithm %s. Allowed values are %s", c.LoadBalance, GlobalLoadBalancingAlgorithms)
+		return fmt.Errorf("Invalid global load balancing algorithm %s. Allowed values are %s", c.LoadBalance, GlobalLoadBalancingStrategies)
 	}
 
 	validAZPref := false
--- a/src/code.cloudfoundry.org/gorouter/config/config.go
+++ b/src/code.cloudfoundry.org/gorouter/config/config.go
@@ -39,8 +39,8 @@ const (
 )
 
 var (
-	GlobalLoadBalancingAlgorithms   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC} // These strategies can be set globally via config
-	LoadBalancingStrategies         = append(GlobalLoadBalancingAlgorithms, LOAD_BALANCE_HB)
+	GlobalLoadBalancingAlgorithms   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}             // These strategies can be set globally via routing-release config
+	LoadBalancingStrategies         = append(GlobalLoadBalancingAlgorithms, LOAD_BALANCE_HB) // These strategies can be set for individual routes (per-route options)
 	AZPreferences                   = []string{AZ_PREF_NONE, AZ_PREF_LOCAL}
 	AllowedShardingModes            = []string{SHARD_ALL, SHARD_SEGMENTS, SHARD_SHARED_AND_SEGMENTS}
 	AllowedForwardedClientCertModes = []string{ALWAYS_FORWARD, FORWARD, SANITIZE_SET}
--- a/src/code.cloudfoundry.org/gorouter/config/config.go
+++ b/src/code.cloudfoundry.org/gorouter/config/config.go
@@ -24,6 +24,7 @@ import (
 const (
 	LOAD_BALANCE_RR           string = "round-robin"
 	LOAD_BALANCE_LC           string = "least-connection"
+	LOAD_BALANCE_HB           string = "hash"
 	AZ_PREF_NONE              string = "none"
 	AZ_PREF_LOCAL             string = "locally-optimistic"
 	SHARD_ALL                 string = "all"
@@ -38,7 +39,8 @@ const (
 )
 
 var (
-	LoadBalancingStrategies         = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}
+	GlobalLoadBalancingAlgorithms   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC} // These strategies can be set globally via config
+	LoadBalancingStrategies         = append(GlobalLoadBalancingAlgorithms, LOAD_BALANCE_HB)
 	AZPreferences                   = []string{AZ_PREF_NONE, AZ_PREF_LOCAL}
 	AllowedShardingModes            = []string{SHARD_ALL, SHARD_SEGMENTS, SHARD_SHARED_AND_SEGMENTS}
 	AllowedForwardedClientCertModes = []string{ALWAYS_FORWARD, FORWARD, SANITIZE_SET}
@@ -595,6 +597,10 @@ func DefaultConfig() (*Config, error) {
 	return &c, nil
 }
 
+func IsGlobalLoadBalancingAlgorithmValid(lbAlgo string) bool {
+	return slices.Contains(GlobalLoadBalancingAlgorithms, lbAlgo)
+}
+
 func IsLoadBalancingAlgorithmValid(lbAlgo string) bool {
 	return slices.Contains(LoadBalancingStrategies, lbAlgo)
 }
@@ -755,8 +761,8 @@ func (c *Config) Process() error {
 		c.RouteServiceEnabled = true
 	}
 
-	if !IsLoadBalancingAlgorithmValid(c.LoadBalance) {
-		return fmt.Errorf("Invalid load balancing algorithm %s. Allowed values are %s", c.LoadBalance, LoadBalancingStrategies)
+	if !IsGlobalLoadBalancingAlgorithmValid(c.LoadBalance) {
+		return fmt.Errorf("Invalid global load balancing algorithm %s. Allowed values are %s", c.LoadBalance, GlobalLoadBalancingAlgorithms)
 	}
 
 	validAZPref := false
--- a/src/code.cloudfoundry.org/gorouter/config/config.go
+++ b/src/code.cloudfoundry.org/gorouter/config/config.go
@@ -39,8 +39,8 @@ const (
 )
 
 var (
-	GlobalLoadBalancingAlgorithms   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}             // These strategies can be set globally via routing-release config
-	LoadBalancingStrategies         = append(GlobalLoadBalancingAlgorithms, LOAD_BALANCE_HB) // These strategies can be set for individual routes (per-route options)
+	GlobalLoadBalancingAlgorithms   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}
+	LoadBalancingStrategies         = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC, LOAD_BALANCE_HB}
 	AZPreferences                   = []string{AZ_PREF_NONE, AZ_PREF_LOCAL}
 	AllowedShardingModes            = []string{SHARD_ALL, SHARD_SEGMENTS, SHARD_SHARED_AND_SEGMENTS}
 	AllowedForwardedClientCertModes = []string{ALWAYS_FORWARD, FORWARD, SANITIZE_SET}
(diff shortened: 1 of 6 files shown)


src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
```diff
@@ -41,7 +41,9 @@ type RegistryMessage struct {
 }
 
 type RegistryMessageOpts struct {
-	LoadBalancingAlgorithm string `json:"loadbalancing"`
+	LoadBalancingAlgorithm string  `json:"loadbalancing"`
+	HashHeader             string  `json:"hash_header"`
```
maxmoehl: I faintly remember that there was the idea to hash things other than headers, would this then be an additional config option?

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -74,6 +74,21 @@ type ProxyRoundTripper interface {
 	CancelRequest(*http.Request)
 }
 
+type HashRoutingProperties struct {
+	Header        string
+	BalanceFactor float64
```
maxmoehl: Can you please add a comment explaining what this value means and what the range of valid values is?

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
maxmoehl, line 1: Wow, this diff is all over the place, thanks GitHub!

<sub>(you might be partially at fault by moving some code around but the `git` CLI properly detects that)</sub>

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -74,6 +74,21 @@ type ProxyRoundTripper interface {
 	CancelRequest(*http.Request)
 }
 
+type HashRoutingProperties struct {
```
maxmoehl: I'm not sure if I agree with splitting this into a separate struct. If we were to do this it would probably be more consistent to have a *Options struct to keep some resemblance to the structure of the message. Alternative is to put the fields directly into the Endpoint.

Edit: especially since the EndpointOpts does exactly that and it would remove the risk of a bunch of nil pointer panics.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -221,6 +241,16 @@ func NewEndpoint(opts *EndpointOpts) *Endpoint {
 		UpdatedAt:              opts.UpdatedAt,
 		LoadBalancingAlgorithm: opts.LoadBalancingAlgorithm,
 	}
+
+	// TODO: Log debug? warning when HashHeader is set but LoadBalancingAlgorithm is not LOAD_BALANCE_HB?
```
maxmoehl: Well, this would be a bug, no? The idea is that these cases are prevented by CC. So IMO we could add a warning level log that says something like:

> BUG: Setting lb-algo to hash should not be possible without a header, please report this to the routing-release GitHub project.

src/code.cloudfoundry.org/gorouter/route/pool_test.go
```diff
@@ -232,134 +296,158 @@ var _ = Describe("EndpointPool", func() {
...
+					LoadBalancingAlgorithm: config.LOAD_BALANCE_RR,
+				})
+				iterator := poolWithLBAlgoLC.Endpoints(logger.Logger, "", false, "none", "az")
+				Expect(iterator).To(BeAssignableToTypeOf(&route.RoundRobin{}))
+				Eventually(logger).Should(gbytes.Say(`endpoint-iterator-with-round-robin-lb-algo`))
 			})
-			iterator := poolWithLBAlgoLC.Endpoints(logger.Logger, "", false, "none", "az")
-			Expect(iterator).To(BeAssignableToTypeOf(&route.RoundRobin{}))
-			Eventually(logger).Should(gbytes.Say(`endpoint-iterator-with-round-robin-lb-algo`))
+
```
maxmoehl, line 339: This seems unnecessary?

src/code.cloudfoundry.org/gorouter/route/pool_test.go
```diff
@@ -232,134 +296,158 @@ var _ = Describe("EndpointPool", func() {
...
+				//	})
+				//	iterator := poolWithLBAlgoHB.Endpoints(logger.Logger, "", false, "none", "az")
+				//	Expect(iterator).To(BeAssignableToTypeOf(&route.HashBased{}))
+				//	Eventually(logger).Should(gbytes.Say(`endpoint-iterator-with-round-robin-lb-algo`))
+			})
 
-			pool.Put(updatedEndpoint)
-			Expect(pool.LoadBalancingAlgorithm).To(Equal(config.LOAD_BALANCE_RR))
+			It("overwrites the load balancing algorithm and HashRoutingProperties of the endpoint and pool", func() {
+				//	TODO
```
maxmoehl: 😄 I'm guessing this is intentional.

src/code.cloudfoundry.org/gorouter/route/pool.go
b1tamara: Hash-based routing may cover a small percent of use cases, and the properties hashHeader and HashBalanceFactor will not exist for the most Endpoints. Using a separate struct, we can check it simply, if it is hash-based routing or not.

src/code.cloudfoundry.org/gorouter/route/pool_test.go
b1tamara: What exactly is unnecessary?

src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
a18e: We are implementing the [RFC in its current state](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0042-hash-based-routing.md#gorouter), which focuses on the Header option.

We are having some discussions about adding more functionality (cookies, URL Params) later on (maybe even before going live with the feature, but not now).

maxmoehl: Should've read that before reviewing the code, sorry 🙈

src/code.cloudfoundry.org/gorouter/route/pool_test.go
maxmoehl: The diff view fooled me here as well, this should be fine.

src/code.cloudfoundry.org/gorouter/route/pool.go
a18e: Done in c9cbf056a28140948a50d468a02e3547dd38db89

src/code.cloudfoundry.org/gorouter/route/pool.go
a18e: Yep, it would be a bug. But this check was in the wrong place anyways.
We're checking if the loadbalancing algorithm is valid when writing it [from the endpoint to the endpoint pool](https://github.com/cloudfoundry/routing-release/blob/c9cbf056a28140948a50d468a02e3547dd38db89/src/code.cloudfoundry.org/gorouter/route/pool.go#L578-L591).
This will be included in a followup PR. (endpoint pool isn't in scope for the current PR)

src/code.cloudfoundry.org/gorouter/route/pool_test.go
a18e: okok, removed it... Will also be included in followup (maybe even w/o TODO 🤯 )

src/code.cloudfoundry.org/gorouter/route/pool.go
maxmoehl: Checking which algorithm is used is done via `LoadBalancingAlgorithm`, no? And if it is set to `hash` the rest of the code can assume that the required properties were checked during the endpoint creation and can safely be used.

src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -38,7 +39,8 @@ const (
 )
 
 var (
-	LoadBalancingStrategies         = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}
+	GlobalLoadBalancingAlgorithms   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}             // These strategies can be set globally via routing-release config
+	LoadBalancingStrategies         = append(GlobalLoadBalancingAlgorithms, LOAD_BALANCE_HB) // These strategies can be set for individual routes (per-route options)
```
maxmoehl: I would prefer two explicit lists:

```suggestion
	GlobalLoadBalancingAlgorithms   = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}
	LoadBalancingStrategies         = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC, LOAD_BALANCE_HB}
```

tying together unrelated lists has bitten us in the pasts with the error checking. When adding a new algorithm the developer should make the conscious choice to add it to either, or, or both.

(I also think the names speak for themselves, removed the comment in the suggestion)

b1tamara: I like your idea.

maxmoehl (dismissed):

maxmoehl:
Not merging yet as we still have to run the test suite to make sure we didn't break anything.

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
```diff
@@ -544,18 +620,26 @@ var _ = Describe("Subscriber", func() {
...
+				var expectedLBAlgo = "hash"
+
+				var msg = mbus.RegistryMessage{
 					Host:     "host",
 					App:      "app",
 					Protocol: "http2",
 					Uris:     []route.Uri{"test.example.com"},
+					Options: mbus.RegistryMessageOpts{
+						LoadBalancingAlgorithm: expectedLBAlgo,
+						HashHeader:             "X-Header",
```
Soha-Albaghdady: You forgot to adapt the test after your last comment
HashHeader -> HashHeaderName
HashBalance -> HashBalanceFactor

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
```diff
@@ -565,57 +649,80 @@ var _ = Describe("Subscriber", func() {
...
 				_, originalEndpoint := registry.RegisterArgsForCall(0)
 				expectedEndpoint := route.NewEndpoint(&route.EndpointOpts{
-					Host:     "host",
-					AppId:    "app",
-					Protocol: "http1",
+					Host:                   "host",
+					AppId:                  "app",
+					Protocol:               "http2",
+					LoadBalancingAlgorithm: expectedLBAlgo,
+					HashHeader:             "X-Header",
```
Soha-Albaghdady: See the comment above

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
```diff
@@ -565,57 +649,80 @@ var _ = Describe("Subscriber", func() {
...
+
+				Eventually(registry.RegisterCallCount).Should(Equal(1))
+				_, originalEndpoint := registry.RegisterArgsForCall(0)
+				expectedEndpoint := route.NewEndpoint(&route.EndpointOpts{
+					Host:                   "host",
+					AppId:                  "app",
+					Protocol:               "http2",
+					LoadBalancingAlgorithm: expectedLBAlgo,
+				})
+				Expect(expectedEndpoint.HashRoutingProperties).To(BeNil())
```
Soha-Albaghdady: HashRoutingProperties no longer exists

Soha-Albaghdady (changes requested):

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
```diff
@@ -565,57 +649,80 @@ var _ = Describe("Subscriber", func() {
...
-			data, err := json.Marshal(msg)
-			Expect(err).NotTo(HaveOccurred())
+				var msg = mbus.RegistryMessage{
+					Host:     "host",
+					App:      "app",
+					Protocol: "http2",
+					Uris:     []route.Uri{"test.example.com"},
+					Options: mbus.RegistryMessageOpts{
+						LoadBalancingAlgorithm: expectedLBAlgo,
+						HashHeader:             "X-Header",
```
Soha-Albaghdady: HashHeader -> HashHeaderName

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
```diff
@@ -565,57 +649,80 @@ var _ = Describe("Subscriber", func() {
...
-				StaleThresholdInSeconds: 120,
-				Tags:                    map[string]string{"key": "value"},
+				Eventually(registry.RegisterCallCount).Should(Equal(1))
+				_, originalEndpoint := registry.RegisterArgsForCall(0)
+				expectedEndpoint := route.NewEndpoint(&route.EndpointOpts{
+					Host:                   "host",
+					AppId:                  "app",
+					Protocol:               "http2",
+					LoadBalancingAlgorithm: expectedLBAlgo,
+					HashHeader:             "X-Header",
```
Soha-Albaghdady: HashHeader -> HashHeaderName
HashBalance -> HashBalanceFactor

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
```diff
@@ -565,57 +649,80 @@ var _ = Describe("Subscriber", func() {
...
+				var expectedLBAlgo = "round-robin"
+
+				var msg = mbus.RegistryMessage{
+					Host:     "host",
+					App:      "app",
+					Protocol: "http2",
+					Uris:     []route.Uri{"test.example.com"},
+					Options: mbus.RegistryMessageOpts{
+						LoadBalancingAlgorithm: expectedLBAlgo,
+						HashHeader:             "X-Header",
```
Soha-Albaghdady: HashHeader -> HashHeaderName

b1tamara: done! Thank you

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
b1tamara: Done!

src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
b1tamara: Done!

Soha-Albaghdady (approved):

b1tamara:
```
~/workspace/routing-release$ ./scripts/test-in-docker.bash
Ginkgo ran 14 suites in 3m33.425684714s
Test Suite Passed
```

maxmoehl (approved):

PR #513, opened 2025-09-11
fix: prevent race conditions in request tracing

maxmoehl:

Summary
---------------

The documentation of net/http/httptrace.ClientTrace clearly states the following:

> Functions may be called concurrently from different goroutines and some
> may be called after the request has completed or failed.

The implementation neglegted this detail and assumed that after RoundTrip returns we can simply reset the timings and only the new timings will be recorded from that point on.

This commit adjusts the round-tripper to clone the request for each try as this is the only way to set the context for a request which carries the trace and its callbacks. For cloning the original context (without the previous trace) is used and a new trace is then attached to the request. It also removes the reset function as it is no longer used.

~I still have to figure out the testing setup as I have failures on my machine both for develop and my branch, keeping this as a draft until I figure this out.~ Was mostly related to running with go1.25, setting it to go1.24 (the version the release is also built with) and running on amd64 natively resolved the issues, the suite passes.

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
@@ -102,9 +102,8 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 	var res *http.Response
 	var endpoint *route.Endpoint
 
-	request := originalRequest.Clone(originalRequest.Context())
-	request, trace := traceRequest(request)
-
+	originalCtx := originalRequest.Context()
+	request := originalRequest.Clone(originalCtx)
 	if request.Body != nil {
 		// Temporarily disable closing of the body while in the RoundTrip function, since
 		// the underlying Transport will close the client request body.
@@ -143,11 +142,22 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 	}
 	triedEndpoints := map[string]bool{}
 
+	// It is safe to use trace inside the loop unconditionally as we set it as
+	// the first thing, but code outside the for loop should check if it's nil
+	// to be on the safe side.
+	var trace *requestTracer
 	for attempt := 1; attempt <= maxAttempts; attempt++ {
 		logger := rt.logger
 
-		// Reset the trace to prepare for new times and prevent old data from polluting our results.
-		trace.Reset()
+		// We have to set the tracing for each round-trip as we otherwise risk
+		// having some asynchronous callback modifying the previous trace even
+		// after we have reset it as the handling of connections and requests
+		// in net/http is performing most of its task concurrently. If we re-use
+		// the previous context and set the trace again this causes us to
+		// accumulate callbacks as new traces are called in addition to the
+		// existing ones which could cause issues if a request is retried a lot.
+		request := request.Clone(originalCtx)
+		request, trace = traceRequest(request)
 
 		if reqInfo.RouteServiceURL == nil {
 			// Because this for-loop is 1-indexed, we substract one from the attempt value passed to selectEndpoint,
@@ -337,14 +347,17 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 		rt.combinedReporter.CaptureWebSocketUpdate()
 	}
 
-	// Record the times from the last attempt, but only if it succeeded.
-	reqInfo.DnsStartedAt = trace.DnsStart()
-	reqInfo.DnsFinishedAt = trace.DnsDone()
-	reqInfo.DialStartedAt = trace.DialStart()
-	reqInfo.DialFinishedAt = trace.DialDone()
-	reqInfo.TlsHandshakeStartedAt = trace.TlsStart()
-	reqInfo.TlsHandshakeFinishedAt = trace.TlsDone()
-	reqInfo.LocalAddress = trace.LocalAddr()
+	if trace != nil {
+		// Record the times from the last attempt, but only on success and if we
+		// have a trace.
+		reqInfo.DnsStartedAt = trace.DnsStart()
+		reqInfo.DnsFinishedAt = trace.DnsDone()
+		reqInfo.DialStartedAt = trace.DialStart()
+		reqInfo.DialFinishedAt = trace.DialDone()
+		reqInfo.TlsHandshakeStartedAt = trace.TlsStart()
+		reqInfo.TlsHandshakeFinishedAt = trace.TlsDone()
+		reqInfo.LocalAddress = trace.LocalAddr()
+	}
 
 	if res != nil && endpoint.PrivateInstanceId != "" && !requestSentToRouteService(request) {
 		setupStickySession(
--- a/src/code.cloudfoundry.org/gorouter/integration/retry_test.go
+++ b/src/code.cloudfoundry.org/gorouter/integration/retry_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"strconv"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -54,7 +55,7 @@ var _ = Describe("Retries", func() {
 		})
 
 		It("does not prune the endpoint on context cancelled", func() {
-			conn, err := net.Dial("tcp", fmt.Sprintf("%s:%d", appURL, testState.cfg.Port))
+			conn, err := net.Dial("tcp", net.JoinHostPort(appURL, strconv.FormatInt(int64(testState.cfg.Port), 10)))
 			Expect(err).ToNot(HaveOccurred())
 
 			_, err = conn.Write([]byte(fmt.Sprintf("GET / HTTP/1.1\r\nHost: %s\r\n\r\n", appURL)))
--- a/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/trace.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/trace.go
@@ -24,20 +24,6 @@ type requestTracer struct {
 	tlsDone   atomic.Int64
 }
 
-// Reset the trace data. Helpful when performing the same request again.
-func (t *requestTracer) Reset() {
-	t.gotConn.Store(false)
-	t.connReused.Store(false)
-	t.wroteHeaders.Store(false)
-	t.localAddr.Store(nil)
-	t.dnsStart.Store(0)
-	t.dnsDone.Store(0)
-	t.dialStart.Store(0)
-	t.dialDone.Store(0)
-	t.tlsStart.Store(0)
-	t.tlsDone.Store(0)
-}
-
 // GotConn returns true if a connection (TCP + TLS) to the backend was established on the traced request.
 func (t *requestTracer) GotConn() bool {
 	return t.gotConn.Load()
--- a/src/code.cloudfoundry.org/gorouter/router/router_test.go
+++ b/src/code.cloudfoundry.org/gorouter/router/router_test.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"net/http/httptrace"
 	"os"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -721,7 +722,7 @@ var _ = Describe("Router", func() {
 			return appRegistered(registry, app)
 		}).Should(BeTrue())
 
-		conn, err := net.Dial("tcp", fmt.Sprintf("%s:%d", config.Ip, config.Port))
+		conn, err := net.Dial("tcp", net.JoinHostPort(config.Ip, strconv.FormatInt(int64(config.Port), 10)))
 		Expect(err).NotTo(HaveOccurred())
 		defer conn.Close()
 


Dariquest:
LGTM

Dariquest (approved):

a18e (dismissed):

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -143,11 +142,22 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
...
-		trace.Reset()
+		// We have to set the tracing for each round-trip as we otherwise risk
+		// having some asynchronous callback modifying the previous trace even
+		// after we have reset it as the handling of connections and requests
+		// in net/http is performing most of its task concurrently. If we re-use
+		// the previous context and set the trace again this causes us to
+		// accumulate callbacks as new traces are called in addition to the
+		// existing ones which could cause issues if a request is retried a lot.
+		request := request.Clone(originalCtx)
+		request, trace := traceRequest(request)
```
ameowlia: aren't you shadowing the trace var here in the for loop with the `:=`. Won't it always be nil outside of the loop?

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -337,14 +347,17 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
...
+	if trace != nil {
+		// Record the times from the last attempt, but only on success and if we
+		// have a trace.
+		reqInfo.DnsStartedAt = trace.DnsStart()
+		reqInfo.DnsFinishedAt = trace.DnsDone()
+		reqInfo.DialStartedAt = trace.DialStart()
+		reqInfo.DialFinishedAt = trace.DialDone()
+		reqInfo.TlsHandshakeStartedAt = trace.TlsStart()
+		reqInfo.TlsHandshakeFinishedAt = trace.TlsDone()
+		reqInfo.LocalAddress = trace.LocalAddr()
```
ameowlia, line 359: How does the access log handler deal with none of these reqInfo variables being set? Do we have tests for that?

ameowlia (changes requested):
Please ping me on slack when you address my comments. Ty!

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
maxmoehl: Oh boy, that would've been nasty 😅

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
maxmoehl: They all check for zero times:

https://github.com/cloudfoundry/routing-release/blob/df741e6d43a91a2f3c7a60679c8500e4a718ef84/src/code.cloudfoundry.org/gorouter/accesslog/schema/access_log_record.go#L140-L159

and this should be the corresponding test case:

https://github.com/cloudfoundry/routing-release/blob/df741e6d43a91a2f3c7a60679c8500e4a718ef84/src/code.cloudfoundry.org/gorouter/accesslog/schema/access_log_record_test.go#L374-L389

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
maxmoehl: I removed the `:`.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -143,11 +142,22 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
...
-		// Reset the trace to prepare for new times and prevent old data from polluting our results.
-		trace.Reset()
+		// We have to set the tracing for each round-trip as we otherwise risk
+		// having some asynchronous callback modifying the previous trace even
+		// after we have reset it as the handling of connections and requests
+		// in net/http is performing most of its task concurrently. If we re-use
+		// the previous context and set the trace again this causes us to
+		// accumulate callbacks as new traces are called in addition to the
+		// existing ones which could cause issues if a request is retried a lot.
+		request := request.Clone(originalCtx)
```
geofffranks, line 159: this request is also masking the previous one (in multiple ways).
Why do we do this at the beginning of the loop, and ignore the originally cloned request instead of re-clone at the end if another retry is needed?

There's logic in between the first clone and the in-loop clone that involves the ofirst clone'd request, which includes reconfiguring the request body to an io.NoOpCloser so it can't be closed until the request is done. However those changes aren't propagated inside this re-clone, and the clone is using the original request that didn't have those changes applied. So wouldn't a retry possibly result in a failure from the req body already being closed, right?

Maybe it would be safer + easier to understand if the request cloning + tracing stuff were pulled out into their own function,  and called subsequent times lower down in if the request needs to be retried?

maxmoehl: I was assuming that the first run would call the `Clone` on the "outer" request and subsequent iterations would call `Clone` on the request from the previous iteration. Well, turns out that's not the case (explanation below):

```
(dlv) c
Running Suite: RoundTripper Suite - /Users/D074009/oss/cf/routing-release/src/code.cloudfoundry.org/gorouter/proxy/round_tripper
================================================================================================================================
Random Seed: 1758647656

Will run 1 of 140 specs
SSSS> [Breakpoint 1] code.cloudfoundry.org/gorouter/proxy/round_tripper.(*roundTripper).RoundTrip() ./proxy_round_tripper.go:148 (hits goroutine(37):1 total:1) (PC: 0x103412618)
   143:         triedEndpoints := map[string]bool{}
   144:
   145:         // It is safe to use trace inside the loop unconditionally as we set it as
   146:         // the first thing, but code outside the for loop should check if it's nil
   147:         // to be on the safe side.
=> 148:         var trace *requestTracer
   149:         for attempt := 1; attempt <= maxAttempts; attempt++ {
   150:                 logger := rt.logger
   151:
   152:                 // We have to set the tracing for each round-trip as we otherwise risk
   153:                 // having some asynchronous callback modifying the previous trace even

(dlv) print request
("*net/http.Request")(0x140002acf00)
*net/http.Request {
        Method: "GET",
        URL: *net/url.URL {
                Scheme: "http",
                Opaque: "/",
                User: *net/url.Userinfo nil,
                Host: "myapp.com",
                Path: "",
                RawPath: "",
                OmitHost: false,
                ForceQuery: false,
                RawQuery: "",
                Fragment: "",
                RawFragment: "",},
        Proto: "HTTP/1.1",
        ProtoMajor: 1,
        ProtoMinor: 1,
        Header: net/http.Header [],
        Body: io.ReadCloser(io.nopCloserWriterTo) {
                Reader: io.Reader(*code.cloudfoundry.org/gorouter/proxy/round_tripper_test.testBody) ...,},
        GetBody: nil,
        ContentLength: 0,
        TransferEncoding: []string len: 0, cap: 0, nil,
        Close: false,
        Host: "myapp.com",
        Form: net/url.Values nil,
        PostForm: net/url.Values nil,
        MultipartForm: *mime/multipart.Form nil,
        Trailer: net/http.Header nil,
        RemoteAddr: "",
        RequestURI: "",
        TLS: *crypto/tls.ConnectionState nil,
        Cancel: <-chan struct {} {},
        Response: *net/http.Response nil,
        Pattern: "",
        ctx: context.Context(*context.valueCtx) *{
                Context: context.Context(context.backgroundCtx) *(*context.Context)(0x14000229950),
                key: (unreadable could not resolve interface type),
                val: interface {}(*code.cloudfoundry.org/gorouter/handlers.RequestInfo) ...,},
        pat: *net/http.pattern nil,
        matches: []string len: 0, cap: 0, nil,
        otherValues: map[string]string nil,}

(dlv) c
> [Breakpoint 2] code.cloudfoundry.org/gorouter/proxy/round_tripper.(*roundTripper).RoundTrip() ./proxy_round_tripper.go:150 (hits goroutine(37):1 total:1) (PC: 0x1034126bc)
   145:         // It is safe to use trace inside the loop unconditionally as we set it as
   146:         // the first thing, but code outside the for loop should check if it's nil
   147:         // to be on the safe side.
   148:         var trace *requestTracer
   149:         for attempt := 1; attempt <= maxAttempts; attempt++ {
=> 150:                 logger := rt.logger
   151:
   152:                 // We have to set the tracing for each round-trip as we otherwise risk
   153:                 // having some asynchronous callback modifying the previous trace even
   154:                 // after we have reset it as the handling of connections and requests
   155:                 // in net/http is performing most of its task concurrently. If we re-use

(dlv) print attempt
1

(dlv) print request
("*net/http.Request")(0x140002acf00)
*net/http.Request {
        Method: "GET",
        URL: *net/url.URL {
                Scheme: "http",
                Opaque: "/",
                User: *net/url.Userinfo nil,
                Host: "myapp.com",
                Path: "",
                RawPath: "",
                OmitHost: false,
                ForceQuery: false,
                RawQuery: "",
                Fragment: "",
                RawFragment: "",},
        Proto: "HTTP/1.1",
        ProtoMajor: 1,
        ProtoMinor: 1,
        Header: net/http.Header [],
        Body: io.ReadCloser(io.nopCloserWriterTo) {
                Reader: io.Reader(*code.cloudfoundry.org/gorouter/proxy/round_tripper_test.testBody) ...,},
        GetBody: nil,
        ContentLength: 0,
        TransferEncoding: []string len: 0, cap: 0, nil,
        Close: false,
        Host: "myapp.com",
        Form: net/url.Values nil,
        PostForm: net/url.Values nil,
        MultipartForm: *mime/multipart.Form nil,
        Trailer: net/http.Header nil,
        RemoteAddr: "",
        RequestURI: "",
        TLS: *crypto/tls.ConnectionState nil,
        Cancel: <-chan struct {} {},
        Response: *net/http.Response nil,
        Pattern: "",
        ctx: context.Context(*context.valueCtx) *{
                Context: context.Context(context.backgroundCtx) *(*context.Context)(0x14000229950),
                key: (unreadable could not resolve interface type),
                val: interface {}(*code.cloudfoundry.org/gorouter/handlers.RequestInfo) ...,},
        pat: *net/http.pattern nil,
        matches: []string len: 0, cap: 0, nil,
        otherValues: map[string]string nil,}

(dlv) c
> [Breakpoint 3] code.cloudfoundry.org/gorouter/proxy/round_tripper.(*roundTripper).RoundTrip() ./proxy_round_tripper.go:162 (hits goroutine(37):1 total:1) (PC: 0x103412718)
   157:                 // accumulate callbacks as new traces are called in addition to the
   158:                 // existing ones which could cause issues if a request is retried a lot.
   159:                 request := request.Clone(originalCtx)
   160:                 request, trace = traceRequest(request)
   161:
=> 162:                 if reqInfo.RouteServiceURL == nil {
   163:                         // Because this for-loop is 1-indexed, we substract one from the attempt value passed to selectEndpoint,
   164:                         // which expects a 0-indexed value
   165:                         endpoint, selectEndpointErr = rt.selectEndpoint(iter, attempt-1)
   166:
   167:                         if attempt > 1 {

(dlv) print request
("*net/http.Request")(0x140002ad180)

(dlv) c
> [Breakpoint 2] code.cloudfoundry.org/gorouter/proxy/round_tripper.(*roundTripper).RoundTrip() ./proxy_round_tripper.go:150 (hits goroutine(37):2 total:2) (PC: 0x1034126bc)
   145:         // It is safe to use trace inside the loop unconditionally as we set it as
   146:         // the first thing, but code outside the for loop should check if it's nil
   147:         // to be on the safe side.
   148:         var trace *requestTracer
   149:         for attempt := 1; attempt <= maxAttempts; attempt++ {
=> 150:                 logger := rt.logger
   151:
   152:                 // We have to set the tracing for each round-trip as we otherwise risk
   153:                 // having some asynchronous callback modifying the previous trace even
   154:                 // after we have reset it as the handling of connections and requests
   155:                 // in net/http is performing most of its task concurrently. If we re-use

(dlv) print attempt
2

(dlv) print request
("*net/http.Request")(0x140002acf00)

(dlv) c
> [Breakpoint 3] code.cloudfoundry.org/gorouter/proxy/round_tripper.(*roundTripper).RoundTrip() ./proxy_round_tripper.go:162 (hits goroutine(37):2 total:2) (PC: 0x103412718)
   157:                 // accumulate callbacks as new traces are called in addition to the
   158:                 // existing ones which could cause issues if a request is retried a lot.
   159:                 request := request.Clone(originalCtx)
   160:                 request, trace = traceRequest(request)
   161:
=> 162:                 if reqInfo.RouteServiceURL == nil {
   163:                         // Because this for-loop is 1-indexed, we substract one from the attempt value passed to selectEndpoint,
   164:                         // which expects a 0-indexed value
   165:                         endpoint, selectEndpointErr = rt.selectEndpoint(iter, attempt-1)
   166:
   167:                         if attempt > 1 {

(dlv) print request
("*net/http.Request")(0x140002ad7c0)
```

 In the debug session I stepped through the loop using the retries test case.
* The first `print request` is the outer `request` variable (take note of the address).
* The second print is the `request` inside the first iteration before the clone, same address.
* The third print is after the clone (from here on I'm omitting the struct code) with a different address.
* The fourth is again before the clone in the second iteration.
* The fifth (and last) is the one cloned inside the loop again.

I was expecting that the fourth print would give me the address from the third one - but no, it's the address from the first print so we _always_ start out with a reference to the outer `request` structure.

Due to that we also always have the correct modified body. I'm not arguing that this is in any way easy to understand, and the fact that I needed a debugger to properly reason about this is already a red flag :/

maxmoehl: Here is a minimal reproducer that's easier to understand:

https://go.dev/play/p/mZ7pJgLDT1H

<sub>As a fun side-note, LLMs <a href="https://kagi.com/assistant/bcdc9c54-13e0-4f47-8167-9e93df0f2106">fall into the same trap</a> as me although its description hints at the correct solution.</sub>

geofffranks: I'm not sure that reproduction is the same as what's happening in the code. This is what it looks like is happening when I read it:

https://go.dev/play/p/IwuV4gaekvG

maxmoehl: In the debugger output above we can see that every iteration starts out with `request` pointing to `0x140002acf00` which is a pointer to the struct from the outer function scope as shown in the very first print.

And in the code I'm referencing `request` when cloning inside the loop, not `originalRequest` which would indeed be the wrong behavior.

geofffranks: I'm having a very bad time reading. request.Clone(originalCtx) != cloning originalReq.

geofffranks (approved):

ameowlia (approved):

PR #514, opened 2025-09-24
Keep Logs Enabled when Disabling Envelope Metrics 

hoffmaen:

Summary
---------------
With the implementation of Prometheus metrics and the switches to enable/disable Prometheus and Envelope metrics (`enable_envelope_v1_metrics`, `prometheus.enabled`), we are able to replace the Envelope metrics with Prometheus metrics. With the current implementation, the switch to disable Envelope metrics also disables access logs. With this change, access logs are kept even if the Envelope metrics switch is disabled.  

Backward Compatibility
---------------
Breaking Change? **No**
Setting `enable_envelope_v1_metrics: False` will enable access logs with this change. As the switch has been implemented recently, defaults to `True` and doesn't indicate log control, I consider this change as a fix, rather than a breaking change.

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/cmd/gorouter/main.go
+++ b/src/code.cloudfoundry.org/gorouter/cmd/gorouter/main.go
@@ -276,20 +276,20 @@ func main() {
 	os.Exit(0)
 }
 
-// initializeDropsondeReporter setups metrics via dropsonse if enabled
+// initializeDropsondeReporter initializes logs. It also initializes envelope V1 metrics if enabled
 func initializeDropsondeReporter(prefix string, logger *slog.Logger, c *config.Config) *metrics.Metrics {
-	if !c.EnableEnvelopeV1Metrics {
-		return nil
-	}
 	err := dropsonde.Initialize(c.Logging.MetronAddress, c.Logging.JobName)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


src/code.cloudfoundry.org/gorouter/cmd/gorouter/main.go
```diff
@@ -276,15 +276,15 @@ func main() {
 	os.Exit(0)
 }
 
-// initializeDropsondeReporter setups metrics via dropsonse if enabled
+// initializeDropsondeReporter setups logs, and metrics via dropsonde if enabled
```
plowin:
```suggestion
// initializeDropsondeReporter setups logs and metrics via dropsonde if enabled
```

hoffmaen: The comma is in there on purpose; `initializeDropsondeReporter` always setups logs, and metrics only if enabled.

plowin: Maybe make it more explicit. Both fine with me
```suggestion
// initializeDropsondeReporter setups logs. It also sets up metrics via dropsonde if enabled
```

plowin (dismissed):
LGTM, agree to the breaking change assessment

Dariquest (approved):

Dariquest:
LGTM

plowin (approved):

PR #519, opened 2025-11-11
Implement Hash-Based Routing

hoffmaen:

Summary
---------------

This Pull-Request implements #505.

* Implement hash-based routing in Gorouter
  * Gorouter routes requests to endpoints based on the value of a configured request header
  * Gorouter distributes requests to different predefined application instances, when the current request count to an app instance exceeds the balance factor
  * Gorouter routes requests to the "next instance" in the lookup table on retries
* Follow decisions documented in this [comment](https://github.com/cloudfoundry/routing-release/issues/505#issuecomment-3398166141) 

Backward Compatibility
---------------
Breaking Change? **Yes/No**

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -414,6 +414,17 @@ properties:
   router.backends.max_conns:
     description: "Maximum concurrent TCP connections per backend. When set to 0 there is no limit"
     default: 500
+  router.hash_based_routing.lookup_table_size:
+    description: |
+      Size of the Maglev lookup table used for hash-based load balancing.
+      The size should be approximately chosen based on the expected maximum number of endpoints / application instances per route which use hash-based routing. The provided values are not hard limits but recommended values.
+      Valid values are:
+        - "XS": recommended for up to 10 endpoints per pool
+        - "S": recommended for up to 30 endpoints per pool
+        - "M": recommended for up to 50 endpoints per pool
+        - "L": recommended for up to. 100 endpoints per pool
+        - "XL": recommended for up to 650 endpoints per pool
+    default: "S"
   router.tracing.enable_zipkin:
     description: "Enables the addition of the X-B3-Trace-Id header to incoming requests. If the header already exists on the incoming request, it will not be overwritten."
     default: false
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -414,17 +414,6 @@ properties:
   router.backends.max_conns:
     description: "Maximum concurrent TCP connections per backend. When set to 0 there is no limit"
     default: 500
-  router.hash_based_routing.lookup_table_size:
-    description: |
-      Size of the Maglev lookup table used for hash-based load balancing.
-      The size should be approximately chosen based on the expected maximum number of endpoints / application instances per route which use hash-based routing. The provided values are not hard limits but recommended values.
-      Valid values are:
-        - "XS": recommended for up to 10 endpoints per pool
-        - "S": recommended for up to 30 endpoints per pool
-        - "M": recommended for up to 50 endpoints per pool
-        - "L": recommended for up to. 100 endpoints per pool
-        - "XL": recommended for up to 650 endpoints per pool
-    default: "S"
   router.tracing.enable_zipkin:
     description: "Enables the addition of the X-B3-Trace-Id header to incoming requests. If the header already exists on the incoming request, it will not be overwritten."
     default: false
--- a/src/code.cloudfoundry.org/gorouter/config/config.go
+++ b/src/code.cloudfoundry.org/gorouter/config/config.go
@@ -45,6 +45,7 @@ var (
 	AllowedShardingModes            = []string{SHARD_ALL, SHARD_SEGMENTS, SHARD_SHARED_AND_SEGMENTS}
 	AllowedForwardedClientCertModes = []string{ALWAYS_FORWARD, FORWARD, SANITIZE_SET}
 	AllowedQueryParmRedactionModes  = []string{REDACT_QUERY_PARMS_NONE, REDACT_QUERY_PARMS_ALL, REDACT_QUERY_PARMS_HASH}
+	HashBasedLookupTableSizes       = []string{"XS", "S", "M", "L", "XL"}
 )
 
 type StringSet map[string]struct{}
@@ -177,6 +178,10 @@ type RouteServiceConfig struct {
 	TLSPem                    `yaml:",inline"` // embed to get cert_chain and private_key for client authentication
 	EnableWebsockets          bool             `yaml:"enable_websockets"`
 	EgressBlocklist           []string         `yaml:"egress_blocklist,omitempty"`
+}
+
+type HashBasedRoutingConfig struct {
+	LookupTableSize string `yaml:"lookup_table_size"`
 }
 
 type LoggingConfig struct {
@@ -457,7 +462,8 @@ type Config struct {
 	Backends                    BackendConfig `yaml:"backends,omitempty"`
 	ExtraHeadersToLog           []string      `yaml:"extra_headers_to_log,omitempty"`
 
-	RouteServiceConfig RouteServiceConfig `yaml:"route_services,omitempty"`
+	RouteServiceConfig RouteServiceConfig     `yaml:"route_services,omitempty"`
+	HashBasedRouting   HashBasedRoutingConfig `yaml:"hash_based_routing,omitempty"`
 
 	TokenFetcherMaxRetries                    uint32        `yaml:"token_fetcher_max_retries,omitempty"`
 	TokenFetcherRetryInterval                 time.Duration `yaml:"token_fetcher_retry_interval,omitempty"`
@@ -569,6 +575,10 @@ var defaultConfig = Config{
 	// Default load balancer values
 	HealthCheckPollInterval: 10 * time.Second,
 	HealthCheckTimeout:      5 * time.Second,
+
+	HashBasedRouting: HashBasedRoutingConfig{
+		LookupTableSize: "S",
+	},
 }
 
 func DefaultConfig() (*Config, error) {
@@ -582,6 +592,10 @@ func IsGlobalLoadBalancingAlgorithmValid(lbAlgo string) bool {
 
 func IsLoadBalancingAlgorithmValid(lbAlgo string) bool {
 	return slices.Contains(LoadBalancingStrategies, lbAlgo)
+}
+
+func IsHashBasedLookupTableSizeValid(size string) bool {
+	return len(size) == 0 || slices.Contains(HashBasedLookupTableSizes, size)
 }
 
 func (c *Config) Process() error {
@@ -794,6 +808,10 @@ func (c *Config) Process() error {
 	}
 	if !validQueryParamRedaction {
 		return fmt.Errorf("Invalid query param redaction mode: %s. Allowed values are %s", c.Logging.RedactQueryParams, AllowedQueryParmRedactionModes)
+	}
+
+	if !IsHashBasedLookupTableSizeValid(c.HashBasedRouting.LookupTableSize) {
+		return fmt.Errorf("Invalid size %s of lookup table for hash-based routing. Allowed values are %s", c.HashBasedRouting.LookupTableSize, HashBasedLookupTableSizes)
 	}
 
 	if err := c.buildCertPool(); err != nil {
--- a/src/code.cloudfoundry.org/gorouter/config/config.go
+++ b/src/code.cloudfoundry.org/gorouter/config/config.go
@@ -457,7 +457,7 @@ type Config struct {
 	Backends                    BackendConfig `yaml:"backends,omitempty"`
 	ExtraHeadersToLog           []string      `yaml:"extra_headers_to_log,omitempty"`
 
-	RouteServiceConfig RouteServiceConfig     `yaml:"route_services,omitempty"`
+	RouteServiceConfig RouteServiceConfig `yaml:"route_services,omitempty"`
 
 	TokenFetcherMaxRetries                    uint32        `yaml:"token_fetcher_max_retries,omitempty"`
 	TokenFetcherRetryInterval                 time.Duration `yaml:"token_fetcher_retry_interval,omitempty"`
@@ -569,7 +569,6 @@ var defaultConfig = Config{
 	// Default load balancer values
 	HealthCheckPollInterval: 10 * time.Second,
 	HealthCheckTimeout:      5 * time.Second,
-
 }
 
 func DefaultConfig() (*Config, error) {
--- a/src/code.cloudfoundry.org/gorouter/config/config.go
+++ b/src/code.cloudfoundry.org/gorouter/config/config.go
@@ -45,7 +45,6 @@ var (
 	AllowedShardingModes            = []string{SHARD_ALL, SHARD_SEGMENTS, SHARD_SHARED_AND_SEGMENTS}
 	AllowedForwardedClientCertModes = []string{ALWAYS_FORWARD, FORWARD, SANITIZE_SET}
 	AllowedQueryParmRedactionModes  = []string{REDACT_QUERY_PARMS_NONE, REDACT_QUERY_PARMS_ALL, REDACT_QUERY_PARMS_HASH}
-	HashBasedLookupTableSizes       = []string{"XS", "S", "M", "L", "XL"}
 )
 
 type StringSet map[string]struct{}
@@ -178,10 +177,6 @@ type RouteServiceConfig struct {
 	TLSPem                    `yaml:",inline"` // embed to get cert_chain and private_key for client authentication
 	EnableWebsockets          bool             `yaml:"enable_websockets"`
 	EgressBlocklist           []string         `yaml:"egress_blocklist,omitempty"`
-}
-
-type HashBasedRoutingConfig struct {
-	LookupTableSize string `yaml:"lookup_table_size"`
 }
 
 type LoggingConfig struct {
@@ -463,7 +458,6 @@ type Config struct {
 	ExtraHeadersToLog           []string      `yaml:"extra_headers_to_log,omitempty"`
 
 	RouteServiceConfig RouteServiceConfig     `yaml:"route_services,omitempty"`
-	HashBasedRouting   HashBasedRoutingConfig `yaml:"hash_based_routing,omitempty"`
 
 	TokenFetcherMaxRetries                    uint32        `yaml:"token_fetcher_max_retries,omitempty"`
 	TokenFetcherRetryInterval                 time.Duration `yaml:"token_fetcher_retry_interval,omitempty"`
@@ -576,9 +570,6 @@ var defaultConfig = Config{
 	HealthCheckPollInterval: 10 * time.Second,
 	HealthCheckTimeout:      5 * time.Second,
 
-	HashBasedRouting: HashBasedRoutingConfig{
-		LookupTableSize: "S",
-	},
 }
 
 func DefaultConfig() (*Config, error) {
@@ -592,10 +583,6 @@ func IsGlobalLoadBalancingAlgorithmValid(lbAlgo string) bool {
 
 func IsLoadBalancingAlgorithmValid(lbAlgo string) bool {
 	return slices.Contains(LoadBalancingStrategies, lbAlgo)
-}
-
-func IsHashBasedLookupTableSizeValid(size string) bool {
-	return len(size) == 0 || slices.Contains(HashBasedLookupTableSizes, size)
 }
 
 func (c *Config) Process() error {
@@ -808,10 +795,6 @@ func (c *Config) Process() error {
 	}
 	if !validQueryParamRedaction {
 		return fmt.Errorf("Invalid query param redaction mode: %s. Allowed values are %s", c.Logging.RedactQueryParams, AllowedQueryParmRedactionModes)
-	}
-
-	if !IsHashBasedLookupTableSizeValid(c.HashBasedRouting.LookupTableSize) {
-		return fmt.Errorf("Invalid size %s of lookup table for hash-based routing. Allowed values are %s", c.HashBasedRouting.LookupTableSize, HashBasedLookupTableSizes)
 	}
 
 	if err := c.buildCertPool(); err != nil {
--- a/src/code.cloudfoundry.org/gorouter/handlers/helpers.go
+++ b/src/code.cloudfoundry.org/gorouter/handlers/helpers.go
@@ -63,17 +63,17 @@ func upgradeHeader(request *http.Request) string {
 	return ""
 }
 
-func EndpointIteratorForRequest(logger *slog.Logger, request *http.Request, stickySessionCookieNames config.StringSet, authNegotiateSticky bool, locallyOptimistic bool, az string, globalLB string) (route.EndpointIterator, error) {
+func EndpointIteratorForRequest(logger *slog.Logger, request *http.Request, stickySessionCookieNames config.StringSet, authNegotiateSticky bool, locallyOptimistic bool, az string, globalRoutingAlgo string) (route.EndpointIterator, error) {
 	reqInfo, err := ContextRequestInfo(request)
 	if err != nil {
 		return nil, fmt.Errorf("could not find reqInfo in context")
 	}
 	stickyEndpointID, mustBeSticky := GetStickySession(request, stickySessionCookieNames, authNegotiateSticky)
 	routingProperties := route.RoutingProperties{
-		RequestHeaders:    &request.Header,
-		LocallyOptimistic: locallyOptimistic,
-		GlobalLB:          globalLB,
-		AZ:                az,
+		RequestHeaders:         &request.Header,
+		LocallyOptimistic:      locallyOptimistic,
+		GlobalRoutingAlgorithm: globalRoutingAlgo,
+		AZ:                     az,
 	}
 	return reqInfo.RoutePool.Endpoints(logger, stickyEndpointID, mustBeSticky, routingProperties), nil
 }
--- a/src/code.cloudfoundry.org/gorouter/handlers/helpers.go
+++ b/src/code.cloudfoundry.org/gorouter/handlers/helpers.go
@@ -63,13 +63,19 @@ func upgradeHeader(request *http.Request) string {
 	return ""
 }
 
-func EndpointIteratorForRequest(logger *slog.Logger, request *http.Request, stickySessionCookieNames config.StringSet, authNegotiateSticky bool, azPreference string, az string) (route.EndpointIterator, error) {
+func EndpointIteratorForRequest(logger *slog.Logger, request *http.Request, stickySessionCookieNames config.StringSet, authNegotiateSticky bool, locallyOptimistic bool, az string, globalLB string) (route.EndpointIterator, error) {
 	reqInfo, err := ContextRequestInfo(request)
 	if err != nil {
 		return nil, fmt.Errorf("could not find reqInfo in context")
 	}
 	stickyEndpointID, mustBeSticky := GetStickySession(request, stickySessionCookieNames, authNegotiateSticky)
-	return reqInfo.RoutePool.Endpoints(logger, stickyEndpointID, mustBeSticky, azPreference, az), nil
+	routingProperties := route.RoutingProperties{
+		RequestHeaders:    &request.Header,
+		LocallyOptimistic: locallyOptimistic,
+		GlobalLB:          globalLB,
+		AZ:                az,
+	}
+	return reqInfo.RoutePool.Endpoints(logger, stickyEndpointID, mustBeSticky, routingProperties), nil
 }
 
 func GetStickySession(request *http.Request, stickySessionCookieNames config.StringSet, authNegotiateSticky bool) (string, bool) {
--- a/src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
+++ b/src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
@@ -43,7 +43,7 @@ type RegistryMessage struct {
 type RegistryMessageOpts struct {
 	LoadBalancingAlgorithm string  `json:"loadbalancing"`
 	HashHeaderName         string  `json:"hash_header"`
-	HashBalance            float64 `json:"hash_balance"`
+	HashBalance            float64 `json:"hash_balance,string"`
 }
 
 func (rm *RegistryMessage) makeEndpoint(http2Enabled bool) (*route.Endpoint, error) {
--- a/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
@@ -128,7 +128,15 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 
 	stickyEndpointID, mustBeSticky := handlers.GetStickySession(request, rt.config.StickySessionCookieNames, rt.config.StickySessionsForAuthNegotiate)
 	numberOfEndpoints := reqInfo.RoutePool.NumEndpoints()
-	iter := reqInfo.RoutePool.Endpoints(rt.logger, stickyEndpointID, mustBeSticky, rt.config.LoadBalanceAZPreference, rt.config.Zone)
+	locallyOptimistic := rt.config.LoadBalanceAZPreference == config.AZ_PREF_LOCAL
+	routingProperties := route.RoutingProperties{
+		RequestHeaders:    &request.Header,
+		LocallyOptimistic: locallyOptimistic,
+		GlobalLB:          rt.config.LoadBalance,
+		AZ:                rt.config.Zone,
+	}
+
+	iter := reqInfo.RoutePool.Endpoints(rt.logger, stickyEndpointID, mustBeSticky, routingProperties)
 
 	// The selectEndpointErr needs to be tracked separately. If we get an error
 	// while selecting an endpoint we might just have run out of routes. In
--- a/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
@@ -130,10 +130,10 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 	numberOfEndpoints := reqInfo.RoutePool.NumEndpoints()
 	locallyOptimistic := rt.config.LoadBalanceAZPreference == config.AZ_PREF_LOCAL
 	routingProperties := route.RoutingProperties{
-		RequestHeaders:    &request.Header,
-		LocallyOptimistic: locallyOptimistic,
-		GlobalLB:          rt.config.LoadBalance,
-		AZ:                rt.config.Zone,
+		RequestHeaders:         &request.Header,
+		LocallyOptimistic:      locallyOptimistic,
+		GlobalRoutingAlgorithm: rt.config.LoadBalance,
+		AZ:                     rt.config.Zone,
 	}
 
 	iter := reqInfo.RoutePool.Endpoints(rt.logger, stickyEndpointID, mustBeSticky, routingProperties)
--- a/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
@@ -37,7 +37,6 @@ import (
 
 const StickyCookieKey = "JSESSIONID"
 const AZ = "meow-zone"
-const AZPreference = "none"
 
 type testBody struct {
 	bytes.Buffer
@@ -276,10 +275,10 @@ var _ = Describe("ProxyRoundTripper", func() {
 					res, err := proxyRoundTripper.RoundTrip(req)
 					Expect(err).NotTo(HaveOccurred())
 					routingProps := route.RoutingProperties{
-						LocallyOptimistic: false,
-						GlobalLB:          cfg.LoadBalance,
-						AZ:                AZ,
-						RequestHeaders:    &req.Header,
+						LocallyOptimistic:      false,
+						GlobalRoutingAlgorithm: cfg.LoadBalance,
+						AZ:                     AZ,
+						RequestHeaders:         &req.Header,
 					}
 
 					iter := routePool.Endpoints(logger.Logger, "", false, routingProps)
@@ -611,10 +610,10 @@ var _ = Describe("ProxyRoundTripper", func() {
 					})
 
 					routingProps := route.RoutingProperties{
-						LocallyOptimistic: false,
-						GlobalLB:          cfg.LoadBalance,
-						AZ:                AZ,
-						RequestHeaders:    &req.Header,
+						LocallyOptimistic:      false,
+						GlobalRoutingAlgorithm: cfg.LoadBalance,
+						AZ:                     AZ,
+						RequestHeaders:         &req.Header,
 					}
 
 					added := routePool.Put(endpoint)
@@ -2481,46 +2480,43 @@ var _ = Describe("ProxyRoundTripper", func() {
 
 					})
 
-					Context("when there is a JSESSIONID and __VCAP_ID__ set on the request", func() {
-						It("will always route to the instance specified with the __VCAP_ID__ cookie", func() {
+					It("will always route to the instance specified with the __VCAP_ID__ cookie", func() {
 
-							// Generate 20 random values for the hash header, so chance that all go to instanceID1
-							// by accident is 0.33^20
-							for i := 0; i < 20; i++ {
-								randomStr := make([]byte, 8)
-								for j := range randomStr {
-									randomStr[j] = byte('a' + rand.Intn(26))
-								}
-
-								req.Header.Set("X-Hash", string(randomStr))
-								reqInfo, err := handlers.ContextRequestInfo(req)
-								req.AddCookie(&http.Cookie{Name: round_tripper.VcapCookieId, Value: "instanceID1"})
-								req.AddCookie(&http.Cookie{Name: StickyCookieKey, Value: "abc"})
-
-								Expect(err).ToNot(HaveOccurred())
-								reqInfo.RoutePool = routePool
-
-								resp, err := proxyRoundTripper.RoundTrip(req)
-								Expect(err).ToNot(HaveOccurred())
-
-								new_cookies := resp.Cookies()
-								Expect(new_cookies).To(HaveLen(2))
-
-								for _, cookie := range new_cookies {
-									Expect(cookie.Name).To(SatisfyAny(
-										Equal(StickyCookieKey),
-										Equal(round_tripper.VcapCookieId),
-									))
-									if cookie.Name == StickyCookieKey {
-										Expect(cookie.Value).To(Equal("abc"))
-									} else {
-										Expect(cookie.Value).To(Equal("instanceID1"))
-									}
-								}
-
+						// Generate 20 random values for the hash header, so chance that all go to instanceID1
+						// by accident is 0.33^20
+						for i := 0; i < 20; i++ {
+							randomStr := make([]byte, 8)
+							for j := range randomStr {
+								randomStr[j] = byte('a' + rand.Intn(26))
 							}
 
-						})
+							req.Header.Set("X-Hash", string(randomStr))
+							reqInfo, err := handlers.ContextRequestInfo(req)
+							req.AddCookie(&http.Cookie{Name: round_tripper.VcapCookieId, Value: "instanceID1"})
+							req.AddCookie(&http.Cookie{Name: StickyCookieKey, Value: "abc"})
+
+							Expect(err).ToNot(HaveOccurred())
+							reqInfo.RoutePool = routePool
+
+							resp, err := proxyRoundTripper.RoundTrip(req)
+							Expect(err).ToNot(HaveOccurred())
+
+							new_cookies := resp.Cookies()
+							Expect(new_cookies).To(HaveLen(2))
+
+							for _, cookie := range new_cookies {
+								Expect(cookie.Name).To(SatisfyAny(
+									Equal(StickyCookieKey),
+									Equal(round_tripper.VcapCookieId),
+								))
+								if cookie.Name == StickyCookieKey {
+									Expect(cookie.Value).To(Equal("abc"))
+								} else {
+									Expect(cookie.Value).To(Equal("instanceID1"))
+								}
+							}
+
+						}
 					})
 				})
 			})
--- a/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
@@ -2476,7 +2476,7 @@ var _ = Describe("ProxyRoundTripper", func() {
 						Expect(err).ToNot(HaveOccurred())
 
 						cookies = resp.Cookies()
-						Expect(cookies).To(HaveLen(2))
+						Expect(cookies).To(HaveLen(3)) // JSESSIONID + VCAP_ID + VCAP_ID_META
 
 					})
 
@@ -2502,16 +2502,17 @@ var _ = Describe("ProxyRoundTripper", func() {
 							Expect(err).ToNot(HaveOccurred())
 
 							new_cookies := resp.Cookies()
-							Expect(new_cookies).To(HaveLen(2))
+							Expect(new_cookies).To(HaveLen(3))
 
 							for _, cookie := range new_cookies {
 								Expect(cookie.Name).To(SatisfyAny(
 									Equal(StickyCookieKey),
 									Equal(round_tripper.VcapCookieId),
+									Equal(round_tripper.VcapMetaCookieId),
 								))
 								if cookie.Name == StickyCookieKey {
 									Expect(cookie.Value).To(Equal("abc"))
-								} else {
+								} else if cookie.Name == round_tripper.VcapCookieId {
 									Expect(cookie.Value).To(Equal("instanceID1"))
 								}
 							}
--- a/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net"
 	"net/http"
 	"net/http/httptest"
@@ -274,8 +275,14 @@ var _ = Describe("ProxyRoundTripper", func() {
 				It("logs the error and removes offending backend", func() {
 					res, err := proxyRoundTripper.RoundTrip(req)
 					Expect(err).NotTo(HaveOccurred())
+					routingProps := route.RoutingProperties{
+						LocallyOptimistic: false,
+						GlobalLB:          cfg.LoadBalance,
+						AZ:                AZ,
+						RequestHeaders:    &req.Header,
+					}
 
-					iter := routePool.Endpoints(logger.Logger, "", false, AZPreference, AZ)
+					iter := routePool.Endpoints(logger.Logger, "", false, routingProps)
 					ep1 := iter.Next(0)
 					ep2 := iter.Next(1)
 					Expect(ep1.PrivateInstanceId).To(Equal(ep2.PrivateInstanceId))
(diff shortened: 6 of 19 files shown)


src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -127,6 +127,20 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
...
+		} else {
+			headerName := reqInfo.RoutePool.HashRoutingProperties.Header
+			headerValue := request.Header.Get(headerName)
+			if headerValue != "" {
+				iter.(*route.HashBased).HeaderValue = headerValue
+			} else {
+				iter = reqInfo.RoutePool.FallBackToDefaultLoadBalancing(rt.config.LoadBalance, rt.logger, stickyEndpointID, mustBeSticky, rt.config.LoadBalanceAZPreference, rt.config.Zone)
+			}
+		}
+	}
```
maxmoehl: I would prefer if we could keep the round tripper clear of any algorithm specific logic. Could we instead modify the function which returns the endpoint iterator to also accept the request as a whole and pass that into the hash-specific implementation? That way we also keep the door open on extending for other properties to hash on as well as remove the need for ugly casts like `iter.(*route.HashBased).HeaderValue` which can cause nasty bugs.

src/code.cloudfoundry.org/gorouter/route/maglev.go
```diff
@@ -0,0 +1,260 @@
...
+
+/******************************************************************************
+ * Original github.com/kkdai/maglev/maglev.go
+ *
+ * Copyright (c) 2019 Evan Lin (github.com/kkdai)
+ *
+ * This program and the accompanying materials are made available under
+ * the terms of the Apache License, Version 2.0 which is available at
+ * http://www.apache.org/licenses/LICENSE-2.0.
+ ******************************************************************************/
```
maxmoehl: Please use `//` for comments, Apache-2.0 also mandates that you clearly state if you modified the file which it seems like you did.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -320,7 +347,6 @@ func (p *EndpointPool) Put(endpoint *Endpoint) PoolPutResult {
 		// new one.
 		e.Lock()
 		defer e.Unlock()
-
```
maxmoehl, line 323: Please avoid such unrelated changes.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -452,6 +497,23 @@ func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeStic
...
+func (p *EndpointPool) FallBackToDefaultLoadBalancing(defaultLBAlgo string, logger *slog.Logger, initial string, mustBeSticky bool, azPreference string, az string) EndpointIterator {
+	logger.Info("hash-based-routing-header-not-found",
+		slog.String("poolLBAlgorithm", p.LoadBalancingAlgorithm),
+		slog.String("Host", p.host),
+		slog.String("Path", p.contextPath))
+
+	switch defaultLBAlgo {
+	case config.LOAD_BALANCE_LC:
+		logger.Debug("endpoint-iterator-with-least-connection-lb-algo")
+		return NewLeastConnection(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
```
peanball: if you pass in `azPreference` as boolean, like for the other LB algos, you don't need to repeat the condition in-line 3 times.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -433,16 +465,29 @@ func (p *EndpointPool) removeEndpoint(e *endpointElem) {
...
 	p.Update()
+
+	if p.LoadBalancingAlgorithm == config.LOAD_BALANCE_HB {
+		p.HashLookupTable.Remove(e.endpoint.PrivateInstanceId)
+	}
+
 }
 
-func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeSticky bool, azPreference string, az string) EndpointIterator {
+func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeSticky bool, azPreference string, az string, globalLB string, request *http.Request) EndpointIterator {
```
peanball: you're passing in the full request to then dynamically fetch the hash based header. Why not pass an optional reference to `*HashRoutingProperties`? Parsing then can be done at the call site where you have the request anyway.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -433,16 +465,29 @@ func (p *EndpointPool) removeEndpoint(e *endpointElem) {
...
 		logger.Debug("endpoint-iterator-with-least-connection-lb-algo")
 		return NewLeastConnection(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
 	case config.LOAD_BALANCE_RR:
 		logger.Debug("endpoint-iterator-with-round-robin-lb-algo")
 		return NewRoundRobin(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
+	case config.LOAD_BALANCE_HB:
+		if p.HashRoutingProperties == nil || request.Header.Get(p.HashRoutingProperties.Header) == "" {
+			logger.Error("hash-routing-properties-missing", slog.String("host", p.Host()))
+			return p.FallBackToDefaultLoadBalancing(globalLB, logger, initial, mustBeSticky, azPreference, az)
+		}
```
peanball: for the sake of clarity you could pull this out into `func validateHashBasedInputs(poolProperties *HashBasedProperties, requestProperties *HashRoutingValues, logger *slog.Logger) error`
This does not return an error if everything is fine, but logs the error.

Logic then reads: 
* if not validated, return fallback (Early exit)
* otherwise continue doing hash based stuff.

No change to the logic as it is.

Alternatively you could pull out the whole logic and just have it return the proper LB algo implementation (Fallback or hash-based) in the switch block.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -452,6 +497,23 @@ func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeStic
...
+
+	switch defaultLBAlgo {
+	case config.LOAD_BALANCE_LC:
+		logger.Debug("endpoint-iterator-with-least-connection-lb-algo")
+		return NewLeastConnection(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
+	case config.LOAD_BALANCE_RR:
+		logger.Debug("endpoint-iterator-with-round-robin-lb-algo")
+		return NewRoundRobin(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
+	}
+	return NewRoundRobin(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
```
peanball: you could also move this to the `default:` to make it clear in the switch. There is also no debug log for this branch.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -452,6 +497,23 @@ func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeStic
 	}
 }
 
+func (p *EndpointPool) FallBackToDefaultLoadBalancing(defaultLBAlgo string, logger *slog.Logger, initial string, mustBeSticky bool, azPreference string, az string) EndpointIterator {
+	logger.Info("hash-based-routing-header-not-found",
```
peanball: The information could all be in a single log:
* hash based was intended
* routing header not found, and the info which one
* the selected fallback (i.e. `defaultLBAlgo`)

This log could also be in a deferred closure, in case you need to set any data in the `switch`.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -433,16 +465,29 @@ func (p *EndpointPool) removeEndpoint(e *endpointElem) {
...
 		return NewLeastConnection(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
 	case config.LOAD_BALANCE_RR:
 		logger.Debug("endpoint-iterator-with-round-robin-lb-algo")
 		return NewRoundRobin(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
+	case config.LOAD_BALANCE_HB:
+		if p.HashRoutingProperties == nil || request.Header.Get(p.HashRoutingProperties.Header) == "" {
+			logger.Error("hash-routing-properties-missing", slog.String("host", p.Host()))
+			return p.FallBackToDefaultLoadBalancing(globalLB, logger, initial, mustBeSticky, azPreference, az)
+		}
+		headerValue := request.Header.Get(p.HashRoutingProperties.Header)
```
peanball: this would also remove looking up the header twice.

As @maxmoehl mentioned, if we change/extend the logic for hash based routing, it would be good if the data and the decision somewhat central (in a data structure and a function), and not referenced with individual fields in 20 different locations.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+// It returns the same endpoint for the same header value consistently.
+// If the hash lookup fails or the endpoint is not found, nil will be returned.
+func (h *HashBased) Next(attempt int) *Endpoint {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	endpoint := h.findEndpointIfStickySession()
+	if endpoint == nil && h.mustBeSticky {
+		return nil
+	}
```
peanball: can this not be turned around? if `h.mustBeSticky` is false we can skip the whole `findEndpointIfStickySession()` call?

The rest of the non-sticky handling remains as is.

src/code.cloudfoundry.org/gorouter/registry/registry_test.go
```diff
@@ -407,7 +407,7 @@ var _ = Describe("RouteRegistry", func() {
 					Expect(r.NumEndpoints()).To(Equal(1))
 
 					p := r.Lookup("foo.com")
-					Expect(p.Endpoints(logger.Logger, "", false, azPreference, az).Next(0).ModificationTag).To(Equal(modTag))
+					Expect(p.Endpoints(logger.Logger, "", false, azPreference, az, r.DefaultLoadBalancingAlgorithm, nil).Next(0).ModificationTag).To(Equal(modTag))
```
peanball: just looking at these changes that added 2 parameters, and may add more in the future, we might consider having an optional "*RoutingOptions" struct or something that allows passing extra info if needed. This might include the default LB algo, hash based routing config, and extracted values for now.

For other LB algos this could then contain other information. Could be an interface that is then implemented by different structs.

Arguably, the AZ affinity configuration falls into the same category.

src/code.cloudfoundry.org/gorouter/handlers/helpers.go
```diff
@@ -63,13 +63,13 @@ func upgradeHeader(request *http.Request) string {
 	return ""
 }
 
-func EndpointIteratorForRequest(logger *slog.Logger, request *http.Request, stickySessionCookieNames config.StringSet, authNegotiateSticky bool, azPreference string, az string) (route.EndpointIterator, error) {
+func EndpointIteratorForRequest(logger *slog.Logger, request *http.Request, stickySessionCookieNames config.StringSet, authNegotiateSticky bool, azPreference string, az string, globalLB string) (route.EndpointIterator, error) {
 	reqInfo, err := ContextRequestInfo(request)
 	if err != nil {
 		return nil, fmt.Errorf("could not find reqInfo in context")
 	}
 	stickyEndpointID, mustBeSticky := GetStickySession(request, stickySessionCookieNames, authNegotiateSticky)
```
peanball, line 71: similar to `GetStickySession()` there could be a `GetRoutingOptions` that extracts info from the request. See other comments where I complain about passing the request around.

peanball (commented):
First set of comments. I have not looked at the implementation of maglev and the hash based instance determination.

I've focussed on the integration with the existing Gorouter code, route lookup, LB algo handling, etc.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: findEndpointIfStickySession() is intended to be the same as for another EndpointIterators. It is not only about h.mustBeStricky, but also for the case where the initialEndpoint (sticky endpoint) is set.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+		if !h.mustBeSticky {
+			h.logger.Debug("endpoint-missing-choosing-alternate", slog.String("requested-endpoint", h.stickyEndpointID))
+			h.stickyEndpointID = ""
+		}
+	}
+
+	if e != nil {
+		e.RLock()
+		defer e.RUnlock()
+		return e.endpoint
```
peanball: e might not be valid anymore, as it is retrieved without lock. You then acquire a lock and should re-fetch the value to be sure that it's still there.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -433,22 +472,55 @@ func (p *EndpointPool) removeEndpoint(e *endpointElem) {
...
+		if headerValue == "" {
+			return p.createIterator(routingProps.GlobalLB, logger, initial, mustBeSticky, routingProps)
+		}
+		return NewHashBased(logger, p, initial, mustBeSticky, headerValue)
+	}
+
+	return p.createIterator(p.LoadBalancingAlgorithm, logger, initial, mustBeSticky, routingProps)
+}
+
+func (p *EndpointPool) hashBasedInputsValid(header *http.Header, hashProps *HashRoutingProperties, logger *slog.Logger) string {
```
peanball: The name `hashBasedInputsValid` implies a validation (true/false, or returning an error) but just returns a value that _can_ be empty, even though it is an error case.

Better would be to return `(value string, err error)` and check for the error, instead of empty string to have clearer semantics that this is an error case.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -433,22 +472,55 @@ func (p *EndpointPool) removeEndpoint(e *endpointElem) {
...
+
 }
 
-func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeSticky bool, azPreference string, az string) EndpointIterator {
-	switch p.LoadBalancingAlgorithm {
+func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeSticky bool, routingProps RoutingProperties) EndpointIterator {
+	// For hash-based routing, validate inputs and get header value
+	if p.LoadBalancingAlgorithm == config.LOAD_BALANCE_HB {
+		headerValue := p.hashBasedInputsValid(routingProps.RequestHeaders, p.HashRoutingProperties, logger)
+		if headerValue == "" {
```
peanball: Additionally, this condition should still live side by side in `createIterator()`. That is where all other routing algorithms are divided. Having it here "hides" the existence of the hash based routing algo, if you're just looking at `createIterator()`.


My previous comment was about the amount of code (essentially what is now in `hashBasedInputsValid()`) that was in the switch block. A 3 line piece is fine:

```go
// in createIterator():


func (p *EndpointPool) createIterator(lbAlgo string, logger *slog.Logger, initial string, mustBeSticky bool, routingProps RoutingProperties) EndpointIterator {

if (lbAlgo == config.LOAD_BALANCE_HB) {
    headerValue := p.hashBasedInputsValid(routingProps.RequestHeaders, p.HashRoutingProperties, logger)
    if headerValue != "" { // or err != nil { return NewHashBased(...) }
        return NewHashBased(logger, p, initial, mustBeSticky, headerValue)
    }
    // fall back to default LB and let the usual `createIterator` logic create an iterator for the default global LB algo
    lbAlgo = routingProps.GlobalLB
}
 
switch lbAlgo {
  case config.LOAD_BALANCE_LC:
    logger.Debug("endpoint-iterator-with-least-connection-lb-algo")
    return NewLeastConnection(logger, p, initial, mustBeSticky, routingProps.LocallyOptimistic, routingProps.AZ)
  case config.LOAD_BALANCE_RR:
    logger.Debug("endpoint-iterator-with-round-robin-lb-algo")
    return NewRoundRobin(logger, p, initial, mustBeSticky, routingProps.LocallyOptimistic, routingProps.AZ)
default:
  logger.Error("invalid-pool-load-balancing-algorithm",
  slog.String("poolLBAlgorithm", lbAlgo),
  slog.String("Host", p.host),
  slog.String("Path", p.contextPath))
  logger.Debug("endpoint-iterator-with-round-robin-lb-algo")
  return NewRoundRobin(logger, p, initial, mustBeSticky, routingProps.LocallyOptimistic, routingProps.AZ)
  }
}
```

peanball (changes requested):

src/code.cloudfoundry.org/gorouter/route/hash_based.go
peanball: I've overlooked that when `mustBeSticky` is _not_ set, but `stickyEndpointID` is set, it is reset to `""` in `findEndpointIfStickySession()`.

So not a simple 1:1 change.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+	endpoint := e.endpoint
+	return h.IsImbalancedOrOverloaded(endpoint, e.isOverloaded())
+}
+
+func (h *HashBased) IsImbalancedOrOverloaded(endpoint *Endpoint, isEndpointOverloaded bool) bool {
+	avgNumberOfInFlightRequests := h.CalculateAverageLoad()
+	currentInFlightRequestCount := endpoint.Stats.NumberConnections.Count()
+	balanceFactor := h.pool.HashRoutingProperties.BalanceFactor
+
+	if isEndpointOverloaded {
```
peanball: put this first, or at least before computing the average.

If we know that the endpoint is overloaded and exit early anyway, there is no point in computing the avg number and then ignoring it.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+		h.logger.Debug("hash-based-routing-endpoint-overloaded", slog.String("host", h.pool.host), slog.String("endpoint-id", endpoint.PrivateInstanceId), slog.Int64("endpoint-connections", currentInFlightRequestCount))
+		return true
+	}
+
+	// Check if avgNumberOfInFlightRequests is 0 to avoid division by 0
```
peanball: consider exiting early here as well to make the condition simpler to read.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+	return h.IsImbalancedOrOverloaded(endpoint, e.isOverloaded())
+}
+
+func (h *HashBased) IsImbalancedOrOverloaded(endpoint *Endpoint, isEndpointOverloaded bool) bool {
+	avgNumberOfInFlightRequests := h.CalculateAverageLoad()
+	currentInFlightRequestCount := endpoint.Stats.NumberConnections.Count()
+	balanceFactor := h.pool.HashRoutingProperties.BalanceFactor
+
+	if isEndpointOverloaded {
+		h.logger.Debug("hash-based-routing-endpoint-overloaded", slog.String("host", h.pool.host), slog.String("endpoint-id", endpoint.PrivateInstanceId), slog.Int64("endpoint-connections", currentInFlightRequestCount))
```
peanball: as debug log is usually not on, for performance's sake it's better to guard it by checking if debug level is enabled at all in the logger. 

something like
```go
if !h.logger.Enabled(context.Background(), slog.LevelDebug) {
    h.logger.Debug(...)
}
```

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+	return nil
+}
+
+func (h *HashBased) isImbalancedOrOverloaded(e *endpointElem) bool {
+	endpoint := e.endpoint
+	return h.IsImbalancedOrOverloaded(endpoint, e.isOverloaded())
+}
+
+func (h *HashBased) IsImbalancedOrOverloaded(endpoint *Endpoint, isEndpointOverloaded bool) bool {
+	avgNumberOfInFlightRequests := h.CalculateAverageLoad()
```
peanball, line 167: ~~expensive computation with lots of locks, should only be done if absolutely necessary.~~

if we remove the locking to access (atomic) stats counters, it's not that expensive. Nevertheless, do this when needed and bail out early before iterating over all endpoints.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+		endpointElem := h.pool.findById(id)
+		if endpointElem == nil {
+			h.logger.Error("hash-based-routing-failed", slog.String("host", h.pool.host), log.ErrAttr(errors.New("Endpoint not found in pool")), slog.String("endpoint-id", id))
+			currentIndex = (currentIndex + 1) % lookupTableSize
+			continue
+		}
+
+		lastEndpointPrivateId = id
+
+		if h.pool.HashRoutingProperties.BalanceFactor <= 0 || !h.isImbalancedOrOverloaded(endpointElem) {
```
peanball: in which case would / should the balance factor be <= 0, is that the equivalent to the other code fragments' checks for `!(value == 0)`? Please unify checks for values (e.g. those used for division) that you don't want to be 0. Ideally to `<=0`.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+
+		if visitedEndpoints[id] || id == lastEndpointPrivateId {
+			currentIndex = (currentIndex + 1) % lookupTableSize
+			continue
+		}
+		visitedEndpoints[id] = true
+
+		endpointElem := h.pool.findById(id)
+		if endpointElem == nil {
+			h.logger.Error("hash-based-routing-failed", slog.String("host", h.pool.host), log.ErrAttr(errors.New("Endpoint not found in pool")), slog.String("endpoint-id", id))
```
peanball: error string (`"Endpoint not found in pool"`) should not be capitalized.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+
+// CalculateAverageLoad computes the average number of in-flight requests across all endpoints in the pool.
+func (h *HashBased) CalculateAverageLoad() float64 {
+	if len(h.pool.endpoints) == 0 {
+		return 0
+	}
+
+	var currentInFlightRequestCount int64
+	for _, endpointElem := range h.pool.endpoints {
+		endpointElem.RLock()
```
peanball: The existing `isOverloaded()` method (https://github.com/cloudfoundry/routing-release/blob/main/src/code.cloudfoundry.org/gorouter/route/pool.go#L683-L683) does not read lock the endpoint element for accessing its stats. Please note that stats are _not_ the metadata of the endpoint and don't require a read (or write) lock, as all values are atomic already.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+	h.logger.Error("hash-based-routing-failed", slog.String("host", h.pool.host), log.ErrAttr(errors.New("All endpoints are overloaded")))
+	return nil
+}
+
+func (h *HashBased) isImbalancedOrOverloaded(e *endpointElem) bool {
+	endpoint := e.endpoint
+	return h.IsImbalancedOrOverloaded(endpoint, e.isOverloaded())
+}
+
+func (h *HashBased) IsImbalancedOrOverloaded(endpoint *Endpoint, isEndpointOverloaded bool) bool {
```
peanball: The name `IsImbalancedOrOverloaded` is not very clear, also with its distinction of `isImbalancedOrOverloaded`. Is there a use for both?

src/code.cloudfoundry.org/gorouter/route/hash_based.go
```diff
@@ -0,0 +1,251 @@
...
+			}
+			e = nil
+		}
+
+		if e == nil && h.mustBeSticky {
+			h.logger.Debug("endpoint-missing-but-request-must-be-sticky", slog.String("requested-endpoint", h.stickyEndpointID))
+			return nil
+		}
+
+		if !h.mustBeSticky {
```
peanball: this set of conditions could be refactored to be more readable.

The code is also almost 1:1 from leastconnection.go (`Next()`), maybe we can consolidate into one piece of code?

peanball (dismissed):
Some more comments, now on the hash_based.go

src/code.cloudfoundry.org/gorouter/route/pool.go
b1tamara: Reworked

src/code.cloudfoundry.org/gorouter/route/pool.go
b1tamara: Refactored

src/code.cloudfoundry.org/gorouter/registry/registry_test.go
b1tamara: Applied

src/code.cloudfoundry.org/gorouter/route/pool.go
b1tamara: Refactored

src/code.cloudfoundry.org/gorouter/route/pool.go
b1tamara: applied

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: Refactored.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: Applied

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: Done

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: Refactored

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: See commit: https://github.com/cloudfoundry/routing-release/pull/519/commits/6b7daaab57031f9d3bfb7b3d7c089de31e2854cc

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: It is allowed that the hash balance factor is not set at all on the per-route level (equals to balance factor = 0). In this case, the Gorouter should ignore the imbalance.

peanball: Balance factor being optional is fine / understood. My main point was: sometimes this was expressed as `BalanceFactor == 0`, sometimes as `BalanceFactor <= 0`.

There is only this one instance now. Resolving.

src/code.cloudfoundry.org/gorouter/route/roundrobin.go
```diff
@@ -38,43 +37,22 @@ func (r *RoundRobin) Next(attempt int) *Endpoint {
...
 	}
 
-	if e != nil {
-		e.RLock()
-		defer e.RUnlock()
-		r.lastEndpoint = e.endpoint
-		return e.endpoint
+	if r.mustBeSticky {
+		return nil
 	}
```
peanball, line 50: If I understand this logic in `FindStickyEndpoint` correctly, `h.stickyEndpointID` must not be set (i.e. must be `""`) if `h.mustBeSticky` is false.

The check is:
if there's a sticky endpoint ID, find that in the pool.
   - if you find one, and `mustBeSticky` is false, make sure that it's unset and return `nil`. This is actually the last condition in the code, but falls into here. When we don't find the sticky endpoint with the provided ID
   - if `mustBeSticky` is set, 
      - and an endpoint for the sticky endpoint id was found, check if that sticky endpoint is overloaded.
         - if it is, return `nil`
         - if it's not, also return `nil` as the endpoint we found is reset to nil.
      - and no endpoint for the sticky endpoint ID was found, but one should have been there (`mustBeSticky` is true after all), then log a warning and return nil.
If at the end we didn't return `nil`, and have an endpoint available, return that.

This boils down for me to:
1. if `mustBeSticky` is `false`, we must reset `stickyEndpointID` to `""` to disable it.
2. If something must be sticky, we try to find if we can find the sticky endpoint, check if it's overloaded.

So I think we can simplify and pull out the "must be sticky" reset code out, not having to write into a passed variable.


```suggestion
	if !r.mustBeSticky {
		r.initialEndpoint = ""
	}
	e := r.pool.FindStickyEndpoint(r.logger, r.initialEndpoint)
	if e != nil {
		r.lastEndpoint = e
		return e
}
```

accordingly you can then assume that `FindStickyEndpoint` is called when `r.mustBeSticky` is true and eliminate any code that assumes it isn't.

b1tamara: stickyEndpointId is independent of mustBeSticky. You can see [here](https://github.com/cloudfoundry/routing-release/blob/ed22cf361ef75ddb1e10c0d1c78e46d7e15aef8d/src/code.cloudfoundry.org/gorouter/handlers/helpers.go#L75) the properties assignment
If mustBeSticky is false, the function returns the endpoint = stickyEndpointID (if set) found in the endpoint pool.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -578,6 +576,13 @@ func (p *EndpointPool) FindStickyEndpoint(logger *slog.Logger, stickyEndpointID
 	return nil
 }
 
+// logDebugIfEnabled logs a debug message only if debug level is enabled
+func logDebugIfEnabled(logger *slog.Logger, msg string, args ...any) {
+	if logger.Enabled(context.Background(), slog.LevelDebug) {
+		logger.Debug(msg, args...)
+	}
+}
```
peanball, line 572: `msg` is passed as value and is copied every time, even if it's not logged. Passing by reference would get around this.

```suggestion
func logDebugIfEnabled(logger *slog.Logger, msg *string, args ...any) {
	if logger.Enabled(context.Background(), slog.LevelDebug) 	{
		logger.Debug(*msg, args...)
	}
}
```

src/code.cloudfoundry.org/gorouter/route/roundrobin.go
peanball: yes, thanks! I had overlooked the one condition. Thanks for looking into this with me. Now I understand that `mustBeSticky` is related to the Authorization header (NTLM/Kerbebros; https://github.com/cloudfoundry/gorouter/pull/392) and not a general "this is a sticky request".

peanball:
LGTM. Once we have the performance tests, we should move to community review.

hoffmaen:
Good news - we evaluated the performance of gorouter with hash-based routing support.

Our test setup constists of scaled out HAProxy (as a load balancer in front of gorouter), scaled out Diego Cells and a very simple and performant application with 30 instances. This allows us to performance test gorouter via external requests, where gorouter is the bottleneck.

We compared the performance of the latest stable routing-release, and the same release including changes from this Pull-Request. Using `oha` (inofficial successor of `hey`), we put various load patterns on the application and measured throughput, CPU usage of gorouter, and latency.  In summary,
* both versions performed absolutely the same when using round-robin
* we achieved the same performance using hash-based routing and round-robin, when requests are distributed across all instances (achieved by running `oha` with multiple instances, each using a different `hash_header` value)
* various mixed load scenarios (load on a route with round-robin parallel to load on a route with hash) showed no impact on performance

We also explored that requests are routed to next instances as expected, when the max. connection limit of one app instance was reached. The hash_balance factor is effective and allows distribution of requests to next instances before overloading it.

Please feel free to reach out for details on the results!

peanball (commented):
All my comments are addressed, thanks.

Next Steps: Performance numbers and review / feedback from other CF community members.

b1tamara:
During performance tests, we figured out that the Gorouter memory consumption is higher than expected if we keep the [permutation table ](https://github.com/cloudfoundry/routing-release/pull/519/changes#diff-e434cac12ad4eb35ae482bd517229e84615dac4a9415e63649bdbacb4bdfd4a1R70)(table which is used to fill the final maglev lookup table) in memory. We want to investigate this finding further and optimize it. 
Moving PR back to Draft.

jobs/gorouter/spec
```diff
@@ -418,6 +418,18 @@ properties:
...
+    description: |
+      Size of the Maglev lookup table used for hash-based load balancing.
+      This should be a prime number approximately 100x the maximum expected number of endpoints.
+      The provided values are not hard limits but recommended values.
+      Valid values are:
+        - "XS": recommended for up to 10 endpoints in the route pool with hash-based load balancing
+        - "S": recommended for up to 30 endpoints in the route pool with hash-based load balancing
+        - "M": recommended for up to 50 endpoints
+        - "L": recommended for up to. 100 endpoints
+        - "XL": recommended for up to 650 endpoints
```
hoffmaen:
```suggestion
        - "XS": recommended for up to 10 endpoints per pool
        - "S": recommended for up to 30 endpoints per pool
        - "M": recommended for up to 50 endpoints per pool
        - "L": recommended for up to 100 endpoints per pool
        - "XL": recommended for up to 650 endpoints per pool
```

jobs/gorouter/spec
```diff
@@ -418,6 +418,18 @@ properties:
   router.backends.max_conns:
     description: "Maximum concurrent TCP connections per backend. When set to 0 there is no limit"
     default: 500
+  router.hash_based_routing.lookup_table_size:
+    description: |
+      Size of the Maglev lookup table used for hash-based load balancing.
+      This should be a prime number approximately 100x the maximum expected number of endpoints.
+      The provided values are not hard limits but recommended values.
```
hoffmaen:
```suggestion
      The size should be approximately chosen based on the expected maximum number of endpoints / application instances per route which use hash-based routing. The provided values are not hard limits but recommended values.
```

src/code.cloudfoundry.org/gorouter/route/maglev.go
```diff
@@ -0,0 +1,286 @@
...
+)
+
+// Int subtype of lookup table (int16/int32/int64) limits the maximum possible size.
+// Remember to adapt type of lookupTable and permutation properties if you want to use larger sizes.
+var lookupTableSizeNames = map[string]uint64{
+	"XS": 1009,
+	"S":  3001,
+	"M":  5003,
+	"L":  10007,
+	"XL": 32003, // this size requires at least int16
```
hoffmaen:
```suggestion
	"XL": 32003, // this size requires at least int16 for values in `lookupTable`.
```

src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -817,6 +831,10 @@ func (c *Config) Process() error {
 		return fmt.Errorf("Invalid query param redaction mode: %s. Allowed values are %s", c.Logging.RedactQueryParams, AllowedQueryParmRedactionModes)
 	}
 
+	if !IsHashBasedLookupTableSizeValid(c.HashBasedRouting.LookupTableSize) {
+		return fmt.Errorf("Invalid size of lookup table for hash-based routing %s. Allowed values are %s", c.HashBasedRouting.LookupTableSize, HashBasedLookupTableSizes)
```
hoffmaen:
```suggestion
		return fmt.Errorf("Invalid size %s of lookup table for hash-based routing. Allowed values are %s", c.HashBasedRouting.LookupTableSize, HashBasedLookupTableSizes)
```

hoffmaen (commented):
Only documentation / log related comments.

jobs/gorouter/spec
mike-jc: Done

jobs/gorouter/spec
mike-jc: Done

src/code.cloudfoundry.org/gorouter/route/maglev.go
mike-jc: Done

src/code.cloudfoundry.org/gorouter/config/config.go
mike-jc: Done

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: done

maxmoehl: I would leave that to the logger to decide. Where I see this as relevant is when the mere call to `.Debug` would cause expensive up-front allocation / computation. But here we are just passing around arguments. We could probably optimize this a little by not using `slog.String` and instead just passing the args individually. This would defer any additional allocations until after the enabled check, avoiding the potential overhead they may cause. It is a bit less efficient for the logger itself but given that this is debug logging we don't care anyway.

peanball: can you provide an example? Do you mean something like this?:
```go
h.logger.Debug("hash-based-routing-endpoint-overloaded", "host", h.pool.host, "endpoint-id", endpoint.PrivateInstanceId, "endpoint-connections", currentInFlightRequestCount)
```

This would be fine for me too. the call to Debug is then almost as free as with the wrapping condition.

maxmoehl: So we did a small excursion into the internals of go / slog:
* One zero-allocation variant is to always use `Logger.LogAttrs` with at most 5 attributes as the slog package optimizes for that case.
* The other option is to continue using the guard statement as the code path never gets executed in this case.
* The "convenience" option is `Logger.(Debug|Info|...)` but it comes at the cost of one allocation per argument for the boxing of the passed argument into `any`. Though the cost is negligible in the grand scheme of things, unless you do expensive up-front operations like `fmt.Sprintf`.

For readability I prefer either the `LogAttrs` for performance critical code or the `Debug|Info|...` functions where we don't need to squeeze every last ns out of the code. If we were to decide we want to optimize as much as we can the right thing to do would be to refactor the entire code-base to move to `LogAttrs`. I don't think that's needed so I'd just stick with `Debug` without the guard for this use-case, reserving the guard for cases we have truly expensive upfront work.

maxmoehl:
Please make sure to compact the history before merging.

jobs/gorouter/spec
```diff
@@ -406,6 +406,17 @@ properties:
   router.backends.max_conns:
     description: "Maximum concurrent TCP connections per backend. When set to 0 there is no limit"
     default: 500
+  router.hash_based_routing.lookup_table_size:
+    description: |
+      Size of the Maglev lookup table used for hash-based load balancing.
+      The size should be approximately chosen based on the expected maximum number of endpoints / application instances per route which use hash-based routing. The provided values are not hard limits but recommended values.
```
maxmoehl: You say that they are not hard limits, yet you prevent the operator from choosing their own value. Wouldn't an integer and a comment explaining the rough ranges be a better approach here?

src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
```diff
@@ -43,7 +43,7 @@ type RegistryMessage struct {
 type RegistryMessageOpts struct {
 	LoadBalancingAlgorithm string  `json:"loadbalancing"`
 	HashHeaderName         string  `json:"hash_header"`
-	HashBalance            float64 `json:"hash_balance"`
+	HashBalance            float64 `json:"hash_balance,string"`
```
maxmoehl, line 46: Why isn't this just a number in the JSON that we are parsing?

jobs/gorouter/spec
b1tamara: We assumed that it could be difficult for operators to determine the exact value for this table size. 
However, the current "t-shirt size" approach is an MVP for this first pass and may be refined as we gather more data.  We are still discussing internally how to dynamically calculate the maglev lookup table size based on endpoint numbers, rather than defining a static configuration value via the routing release.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: The code has been reworked.

src/code.cloudfoundry.org/gorouter/route/hash_based.go
b1tamara: Aligned to consistently write debugging logs without `h.logger.Enabled(context.Background(), slog.LevelDebug)`

src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
b1tamara: @hoffmaen 
Could you remember why we changed to string?

hoffmaen: Yes. This is because route-emitter sends the value as a string. We didn't want to make changes to route-emitter, and having the value as a string has no practical downsides.
This is also the implementation in [cloud-controller](https://github.com/cloudfoundry/cloud_controller_ng/pull/4746/changes#diff-6191f12e93ba77d1c40a8f0fb0b4ae43df32196338f120342ec70b2dcef5d727R365)

b1tamara:
The PR is ready for review. 

The hash-based routing feature is disabled by default. To enable it, operators should turn on the feature flag using `cf enable-feature-flag hash_based_routing`.

Based on our performance tests ([see all results here](https://github.com/cloudfoundry/routing-release/issues/505#issuecomment-3964793599)), the hash-based routing feature implementation does not influence the performance of the round-robin load balancing when the feature is enabled, but not in use. The performance test results also showed that the performance of the hash-based routing algorithm is comparable to round-robin.

With the current proposal, the operator is required to configure the maglev table size, with a default value as the size is dependent on the expected maximum number of endpoints for the routes using hash-based routing. However, the current proposal provides an MVP for using that feature and will later be enhanced to dynamically calculate the maglev lookup table size based on endpoint numbers.

jobs/gorouter/spec
maxmoehl: I'm just saying that I would put the t-shirt sizes in the docs but allow the operator to put any arbitrary value within the acceptable range. If an operator is unsure, they can just pick the value from the docs, if they know better they can tune to their own likening.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
```diff
@@ -1818,6 +1832,167 @@ var _ = Describe("ProxyRoundTripper", func() {
...
+						if strings.Contains(infoLogs[i], "hash-based-routing-header-value-not-found") {
+							count++
+						}
+					}
+					Expect(count).To(Equal(1))
+					// Verify it still selects an endpoint
+					Expect(reqInfo.RouteEndpoint).ToNot(BeNil())
+				})
+
+				Context("when sticky session cookies (JSESSIONID and VCAP_ID) are on the request", func() {
```
ameowlia, line 2448: This context is basically the same as the context on line 1952. I don't think we need both.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
```diff
@@ -1818,6 +1832,167 @@ var _ = Describe("ProxyRoundTripper", func() {
...
+						resp, err := proxyRoundTripper.RoundTrip(req)
+						Expect(err).ToNot(HaveOccurred())
+
+						cookies = resp.Cookies()
+						Expect(cookies).To(HaveLen(2))
+
+					})
+
+					Context("when there is a JSESSIONID and __VCAP_ID__ set on the request", func() {
+						It("will always route to the instance specified with the __VCAP_ID__ cookie", func() {
```
ameowlia: Why is JSESSIONID included in the context when this test is about __VCAP_ID__ only ?

src/code.cloudfoundry.org/gorouter/route/pool_test.go
```diff
@@ -233,20 +233,27 @@ var _ = Describe("EndpointPool", func() {
...
 			BeforeEach(func() {
 				modTag = models.ModificationTag{}
 				modTag2 = models.ModificationTag{Guid: "abc"}
 				endpoint1 := route.NewEndpoint(&route.EndpointOpts{Host: "1.2.3.4", Port: 5678, ModificationTag: modTag})
 
 				Expect(pool.Put(endpoint1)).To(Equal(route.EndpointAdded))
+
+				routingProps = route.RoutingProperties{
+					LocallyOptimistic: locallyOptimistic,
+					GlobalLB:          config.LOAD_BALANCE_RR,
```
ameowlia: I don't like the name "GlobalLB", it makes me think of the load balancer that sits in front of gorouters. 

How about...
* GlobalRoutingAlgoritm
* DefaultLoadBalancingAlgorithm
..or something like that?

jobs/gorouter/spec
ameowlia: I agree with Max. Assigning t-shirt sizes in the code makes it harder to change those recommendations because of breaking changes. However, updating the docs to reflect new suggestions would be far easier. 

However, I really like that you assigned t-shirt sizes and documented what they were for! Very helpful!

ameowlia (changes requested):

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
b1tamara: @hoffmaen Could you please take a look at this test?

src/code.cloudfoundry.org/gorouter/route/pool_test.go
b1tamara: You are right. I renamed the property

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
b1tamara: Removed the context on 1952 as it is not really needed. An iteration block is fully enough.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
hoffmaen: This is because sessions are only sticky, when BOTH the `JSESSIONID` and the `VCAP_ID` cookie are in the request (excluding the special case for auth negotiation here). When only `VCAP_ID` is in the request, gorouter ignores it.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
hoffmaen: Just FYI: In #534, I am refactoring the test suite for sticky sessions - having the new test case outside the sticky session context is a plus, as we won't have any merge conflicts later.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
ameowlia: ah yes! ty for the explanation!

ameowlia:
Okay, I finally got it deployed and tried it out. Here are my findings...

## 1. Good error when hash_header and hash_balance were provided, but shouldn't have been
This case errored quickly and clearly. I like that it told me exactly which route was having the issue.

```
---
applications:
  - name: proxy-app
...
    routes:
      - route: lc-invalid.apps.pcf.tasonvsphere.com 
        options:
          loadbalancing: round-robin
          hash_header: tenant-id
          hash_balance: 1.25


$ cf push

Pushing app proxy-app to org o / space s as admin...
Applying manifest file /Users/ad008239/workspace/cf-networking-release/src/example-apps/proxy/manifest.yml...
Multiple errors occurred:
- For application 'proxy-app': Route 'lc-invalid.apps.pcf.tasonvsphere.com': Hash header can only be set when loadbalancing is hash
- For application 'proxy-app': Route 'lc-invalid.apps.pcf.tasonvsphere.com': Hash balance can only be set when loadbalancing is hash
FAILED
```

## 2. Less good error when hash headers are missing 
This case took longer to error than in case 1. In this case it rendered the app manifest and started doing some work before it returned the error. The returned error doesn't tell me which route is misconfigured. I know this is a CAPI code path and is not part of this PR, but I wanted to bring it up anyways.

```
cf push
Pushing app proxy-app to org o / space s as admin...
Applying manifest file /Users/ad008239/workspace/cf-networking-release/src/example-apps/proxy/manifest.yml...

Updating with these attributes...
  ---
  applications:
+ - name: proxy-app
    disk-quota: 64M
    memory: 32M
+   default-route: true
+   buildpacks:
+   - go_buildpack
+   env:
+     GOPACKAGENAME: proxy
+     GOVERSION: go1.25
+   routes:
+   - options:
+       hash_balance: 1.25
+       hash_header: meow
+       loadbalancing: hash
+     route: hash.apps.pcf.tasonvsphere.com
+   - options:
+       loadbalancing: hash
+     route: hash-invalid.apps.pcf.tasonvsphere.com
+   - options:
+       hash_balance: 1.25
+       loadbalancing: hash
+     route: hash-no-header.apps.pcf.tasonvsphere.com
+   - options:
+       hash_header: meow
+       loadbalancing: hash
+     route: hash-no-balance.apps.pcf.tasonvsphere.com
+   - options:
+       hash_balance: 0
+       hash_header: meow
+       loadbalancing: hash
+     route: hash-balance-zero.apps.pcf.tasonvsphere.com
+   - options:
+       loadbalancing: least-connection
+     route: lc.apps.pcf.tasonvsphere.com
For application 'proxy-app': Hash header must be present when loadbalancing is set to hash.
FAILED
```

## 3. when hash balance is 0 it is not in the routing table
Is there any reason why it is not included in this case? I think it would be clearer if it was always set there.

```
"hash-balance-zero.apps.pcf.tasonvsphere.com": [
    {
      ...
      "load_balancing_algorithm": "hash",
      "hash_header": "meow"
    }
  ],
    "hash-no-balance.apps.pcf.tasonvsphere.com": [
    {
       ...
      "load_balancing_algorithm": "hash",
      "hash_header": "meow"
    }
  ],
    "hash.apps.pcf.tasonvsphere.com": [
    {
      ...
      "load_balancing_algorithm": "hash",
      "hash_header": "meow",
      "hash_balance": "1.3"
    }
  ],
 ```

## 4. feature works as expected
I was able to follow the examples in the RFC and use the feature!

✅ confirmed that if I don't pass hash_header default algorithm takes over
✅ confirmed that if I do pass hash_header, I get the same route


## 5. don't forget docs!

**Comment by b1tamara** [comment]
@ameowlia 
> 2. Less good error when hash headers are missing

-> https://github.com/cloudfoundry/cloud_controller_ng/issues/4971

> 3. When hash balance is 0 it is not in the routing table

-> Implemented in the [commit](https://github.com/cloudfoundry/routing-release/pull/519/changes/b27531f4448ba3a989c722e7fd0b595ecaaca15e)

> 5. Don't forget docs!

We are working on it: https://github.com/cloudfoundry/docs-dev-guide/pull/569/changes

File: jobs/gorouter/spec (hunk 31 above)
Comment (b1tamara) [reply jobs/gorouter/spec]: @ameowlia, is it ok for you to go ahead with the current solution and keep t-shirt sizing, or would you prefer we rework this part in this PR?

Comment (b1tamara) [reply jobs/gorouter/spec]: Aligned with @maxmoehl  to provide more context on this property.

The Maglev algorithm recommends using a lookup table of size a prime near 100 * endpoint number for optimal hash distribution. 
In the first draft, we implemented the table with a hardcoded default size. Since we did not know the exact numbers and wanted to prevent unnecessary memory use, we decided to make the lookup table size configurable by the operators. At the same time, we needed to explain to operators that they should provide a prime number near 100 times the number of application instances that might use hash-based routing. That was not straightforward. That’s why we chose a t-shirt-size approach to keep the configuration simpler and skip any prime-number validations. 
 
We are currently faced with two options: 
1. Mark this property as experimental, keeping in mind that it may be removed entirely.
2. Avoid exposing the property at all and revert to the hard-coded default value, which is higher than our known use case for hash-based routing.

In both cases, we can later implement a better configuration property or introduce dynamic calculation of the Maglev lookup table.

@ameowlia and @maxmoehl  @plowin WDYT?

Comment (maxmoehl) [reply jobs/gorouter/spec]: One more detail:

The automatic approach would mean that each lookup table starts with a relatively small value as we expect most apps to have less than a dozen endpoints. Once an app grows beyond that, gorouter would have to re-allocate a larger table just for that pool. This allows large apps to function as expected while keeping the overall memory consumption in check.

My personal preference would be to avoid a config knob by marking the entire feature as experimental and start with a hard-coded value that represents a middle ground (e.g. apps with up to 20 endpoints or so). Excessively large apps should't use it in this phase until a proper dynamic logic has been implemented. Alternatively we include the dynamic logic in this PR, but I understand that we should try to get this PR in now as it has been sitting for a really long time already.

**Review by ameowlia — DISMISSED** [review]

File: src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go [hunk 36]
```diff
@@ -2350,6 +2363,165 @@ var _ = Describe("ProxyRoundTripper", func() {
...
+						}
+						resp, err := proxyRoundTripper.RoundTrip(req)
+						Expect(err).ToNot(HaveOccurred())
+
+						cookies = resp.Cookies()
+						Expect(cookies).To(HaveLen(3)) // JSESSIONID + VCAP_ID + VCAP_ID_META
+
+					})
+
+					It("will always route to the instance specified with the __VCAP_ID__ cookie", func() {
```
b1tamara, line 2483: @hoffmaen Please take a look at this test after sticky session behaviour changes

b1tamara (approved):
LGTM

Issue #529, opened 2026-02-17
Fix Session Affinity With Partitioned Session Cookies


hoffmaen:
### Current behavior

There are currently two issues/bugs/unexpected behaviors with Partitioned Session Cookies in Gorouter.

The issues described here originate from the fact that CHIPS (Cookies Having Independent Partitioned State) introduces a new dimension to cookies (see [RFC6265bis](https://www.ietf.org/archive/id/draft-cutler-httpbis-partitioned-cookies-01.html)), allowing multiple cookies with the same name, domain, and path but different partitioned attributes to co-exist. The cookie attributes reside on the client side and are not sent to the server.

## 0. Session Affinity a/k/a Sticky Sessions - Green Path

The following diagram shows how session affinity in Gorouter works in generel:
<img width="324" height="249" alt="Image" src="https://github.com/user-attachments/assets/f2de258d-98f3-4cca-a950-2ce78af0e534" />

## 1. Defaulting to Issue Non-Partitioned Cookie

Gorouter sets an unpartitioned VCAP_ID cookie when the provided one points to a non-existent instance and no new JSESSIONID cookie was issued by the application.

### Context

This issue arises when an application instance sets a partitioned JSESSIONID cookie in the response (causing Gorouter to add a partitioned VCAP_ID cookie), and subsequently, this application instance becomes unavailable.

The next request (containing partitioned VCAP_ID and partitioned JSESSIONID cookies) reaches Gorouter, but Gorouter cannot route to the instance referenced by the VCAP_ID cookie. In this case, Gorouter's default behavior is to route the request to another available instance. When that instance responds without a JSESSIONID cookie, [Gorouter will always set a new unpartitioned VCAP_ID cookie](https://github.com/cloudfoundry/routing-release/blob/f6a73cbaf90bcc23f75c88d5e0c2da759ce58d27/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go#L544C17-L544C28).

### Problem

This leaves the client with:
* A partitioned JSESSIONID cookie
* A partitioned VCAP_ID cookie referencing a non-existent app instance
* An unpartitioned VCAP_ID cookie referencing an existing instance

This creates problems in two scenarios:
* **Non-Third-Party Context**: The client sends all three cookies to Gorouter. Gorouter will either pick the partitioned VCAP_ID cookie first (routing the request to an arbitrary instance) or pick the unpartitioned cookie first (routing it to the instance referenced in that cookie).
[...]

PR #530, opened 2026-02-17
Allow __VCAP_ID__ session cookie to have Max-Age>0

### LINKED ISSUE #429: Copy Max-Age cookie attribute to VCAP_ID unconditionally
### Current Behavior

With the current implementation of session handling, Gorouter sets a `__VCAP_ID__` cookie in the response when a session cookie (by default `JSESSIONID`) is present in the application response. Some cookie attributes, such as the `Expires` timestamp and the `Partitioned` flag, are copied from the session cookie to `__VCAP_ID__`. However, the `Max-Age` attribute is [only copied if its value is negative](https://github.com/cloudfoundry/gorouter/blob/a602bb6a624308630305f7f61689fddd21347dcb/proxy/round_tripper/proxy_round_tripper.go#L481) ([internally represented as `MaxAge < 0` in Go's cookie implementation](https://github.com/golang/go/blob/f65692ea562bf24c21ae46854e98584dd4bcc201/src/net/http/cookie.go#L36)). This is inconsistent behavior, as the `Expires` attribute is always copied, and [`Max-Age` takes precedence over `Expires`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#max-agenumber).

This leads to unexpected and undesired behavior: When an application provides a positive `Max-Age` value for the session cookie, the session cookie expires once `Max-Age` seconds have elapsed, but the `__VCAP_ID__` cookie remains valid as a session cookie (no expiration). In the next request, [Gorouter will not consider the `__VCAP_ID__` cookie because the expired session cookie is missing](https://github.com/cloudfoundry/gorouter/blob/8c9fa99dae1dea4d496175b2baf90e487b2c2cd1/proxy/round_tripper/proxy_round_tripper.go#L445). As a result, the request is routed to an arbitrary application instance, breaking sticky session routing.

The conditional `Max-Age` copying was introduced 10 years ago in a [bugfix](https://github.com/cloudfoundry/gorouter/commit/87a7eeecb882c4854d262a5e4a7686208e22f397) that resolved an issue preventing sticky session re-establishment.

### Desired Behavior

The `Max-Age` cookie attribute should be unconditionally copied from the session cookie to the `__VCAP_ID__` cookie, regardless of whether its value is negative or positive. This ensures that both cookies have synchronized expiration behavior.

### Affected Version

Current `main` branch

hoffmaen:

Summary
---------------

This Pull-Request fixes #429

With this change, we unconditionally copy the `Max-Age` attribute from the `JSESSIONID` cookie to the `__VCAP_ID__` cookie, regardless of whether its value is positive or negative. Previously, the `Max-Age` attribute was only copied when its value was negative.

Backward Compatibility
---------------
Breaking Change? **No**

This Pull-Request impacts the lifecycle of the `__VCAP_ID__` cookie. Previously, when a CF application set a positive `Max-Age` value for the `JSESSIONID` cookie, which defines when the cookie becomes invalid, Gorouter ignored this attribute. This caused the `JSESSIONID` and `__VCAP_ID__` cookies to be invalidated at different times, leading to inconsistency. This Pull-Request fixes this behavior.

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
@@ -1248,6 +1248,124 @@ var _ = Describe("ProxyRoundTripper", func() {
 								Expect(cookies[1].Raw).To(ContainSubstring("Expires=Wed, 01 Jan 2020 01:00:00 GMT; HttpOnly; Secure; SameSite=Strict"))
 							})
 						})
+
+						Context("when the JSESSIONID cookie has a negative value for the MaxAge attribute", func() {
+							BeforeEach(func() {
+								transport.RoundTripStub = func(req *http.Request) (*http.Response, error) {
+									resp := &http.Response{StatusCode: http.StatusTeapot, Header: make(map[string][]string)}
+
+									// Create JSESSIONID with MaxAge=-1, which translates to "Max-Age=0" in the HTTP header (delete cookie immediately)
+									deleteCookie := &http.Cookie{
+										Name:     StickyCookieKey,
+										Value:    "session-to-delete",
+										MaxAge:   -1,
+										Secure:   true,
+										HttpOnly: true,
+										SameSite: http.SameSiteStrictMode,
+									}
+									resp.Header.Add(round_tripper.CookieHeader, deleteCookie.String())
+									return resp, nil
+								}
+							})
+
+							It("copies MaxAge=-1 to VCAP_ID cookie so both cookies are deleted immediately", func() {
+								resp, err := proxyRoundTripper.RoundTrip(req)
+								Expect(err).ToNot(HaveOccurred())
+
+								cookies := resp.Cookies()
+								Expect(cookies).To(HaveLen(2))
+
+								// Verify JSESSIONID has MaxAge=-1
+								Expect(cookies[0].Name).To(Equal(StickyCookieKey))
+								Expect(cookies[0].MaxAge).To(Equal(-1))
+
+								// Verify VCAP_ID also has MaxAge=-1
+								Expect(cookies[1].Name).To(Equal(round_tripper.VcapCookieId))
+								Expect(cookies[1].MaxAge).To(Equal(-1))
+								Expect(cookies[1].Value).To(SatisfyAny(
+									Equal("id-1"),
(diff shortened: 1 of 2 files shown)


src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
```diff
@@ -1209,7 +1209,7 @@ var _ = Describe("ProxyRoundTripper", func() {
 					Expect(removed).To(BeTrue())
 				})
 
-				Context("when there are no cookies on the request", func() {
+				FContext("when there are no cookies on the request", func() {
```
maxmoehl:
```suggestion
				Context("when there are no cookies on the request", func() {
```

maxmoehl:
To give some context I dug through all the code refactoring and subtree merges and found the commit which originally introduced this: 87a7eeecb. It used to be that none of the attributes of the sticky session cookie would be copied but this change introduced the copying of a negative MaxAge property that indicates that the app wants to invalidate a previous session cookie. This seems to be the main target of that commit: allow changing the sticky session by invalidating an old cookie. What this missed (and what is still in there) is that an app might invalidate the old and issue a new cookie in one go which gorouter does not support due to the break in the loop and only ever issuing one VCAP_ID cookie. @hoffmaen and I already discussed this behavior and it's also described in #529.

Edit: should've read the issue which already contains the reference 🥲

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper_test.go
```diff
@@ -1248,6 +1248,124 @@ var _ = Describe("ProxyRoundTripper", func() {
...
+									Equal("id-2")))
+							})
+						})
+
+						Context("when the JSESSIONID cookie has a positive value for the Max-Age attribute", func() {
+							BeforeEach(func() {
+								transport.RoundTripStub = func(req *http.Request) (*http.Response, error) {
+									resp := &http.Response{StatusCode: http.StatusTeapot, Header: make(map[string][]string)}
+
+									// Create JSESSIONID with Max-Age=0 to invalidate it
```
maxmoehl: Some copy-paste issues, please check the comments and make sure they reflect the code below them.

maxmoehl (approved):

PR #531, opened 2026-02-17
Do not update VCAP_ID cookie without JSESSIONID cookie

hoffmaen:

Summary
---------------
This Pull-Request fixes Problem 1 in #529.

This Pull-Request fixes an issue where Gorouter was creating an unpartitioned `__VCAP_ID__` cookie when a client's original sticky session endpoint became unavailable and the backend response didn't contain a `JSESSIONID` cookie. The previous behavior would set a new `__VCAP_ID__` cookie with default attributes whenever the endpoint changed, regardless of whether the application intended to establish a new session. With this change, Gorouter only sets the `__VCAP_ID__` cookie when the backend response explicitly contains a `JSESSIONID` cookie or WWW-Authenticate Negotiate header, ensuring that cookie attributes remain synchronized and applications maintain full control over session lifecycle.

Backward Compatibility
---------------
Breaking Change? **Yes** (but more of a bugfix)

This Pull-Request changes the behavior of __VCAP_ID__ cookie creation in a specific scenario: when a client sends both `JSESSIONID` and `__VCAP_ID__` cookies, the original endpoint is no longer available, and the backend response doesn't contain a new JSESSIONID cookie. Previously, Gorouter would create a new `__VCAP_ID__ `cookie with default (unpartitioned) attributes in this case.
With this change, Gorouter will not create a new `__VCAP_ID__` cookie unless the backend explicitly sets a `JSESSIONID` cookie in the response. This prevents cookie attribute desynchronization where `JSESSIONID` and `__VCAP_ID__` could have conflicting partitioning states. The previous behavior was problematic because applications lost control over the `__VCAP_ID__` cookie attributes (particularly when using partitioned cookies), as Gorouter would override them with unpartitioned defaults. This fix ensures applications can properly control cookie sandboxing through `JSESSIONID` attributes.


docs/03-how-to-use-session-affinity.md
```diff
@@ -7,128 +7,116 @@ tags: [routing-release]
...
+Sticky sessions work correctly with both platform-deployed Cloud Foundry route services and 
+externally deployed route services. The `__VCAP_ID__` cookie is properly forwarded through 
+the route service chain.
 
-Routing Release <0.211.0: If you deploy a reverse proxy to the platform in front of your app, the backend
-app must return the `JSESSIONID` on every response in order to sticky sessions to
-work.
+### What cookie attributes does `__VCAP_ID__` inherit from `JSESSIONID`?
+The `__VCAP_ID__` cookie inherits most attributes from the `JSESSIONID` cookie set by the application. 
+[See implementation](https://github.com/cloudfoundry/gorouter/blob/379860daa83a162ffe0b6039eafb7c8bfa1eaccf/proxy/round_tripper/proxy_round_tripper.go#L312-L355).
```
b1tamara: Can you replace the link? As gorouter repo is archived, use the routing-release repo, e.g. https://github.com/cloudfoundry/routing-release/blob/e9683b43dec98ee211390c28cc8611d6869de801/src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go#L533

b1tamara (approved):
LGTM

hoffmaen:
Putting this back to Draft status, to align of broken scenarios.

hoffmaen:
Closing in lieu of #534 , as this proposal introduces a breaking change which might negatively affect users.

PR #532, opened 2026-02-19
Refactor endpoint selection for sticky sessions

b1tamara:

Summary
---------------
During the implementation of the hash-based routing, we noticed that the code in EndpointIterators for sticky sessions is duplicated. Move the code to a single place and call it from EndpointIterators. 

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/route/pool.go
+++ b/src/code.cloudfoundry.org/gorouter/route/pool.go
@@ -1,7 +1,6 @@
 package route
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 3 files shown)


hoffmaen (approved):

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -464,6 +465,48 @@ func (p *EndpointPool) findById(id string) *endpointElem {
...
 }
 
+// FindStickyEndpoint attempts to find and return a sticky session endpoint.
+// If the endpoint is found and not overloaded, it returns the endpoint.
+// If mustBeSticky is true and the endpoint is missing or overloaded, it returns nil.
+// If mustBeSticky is false and the endpoint is missing or overloaded, it clears the stickyEndpointID and returns nil.
+// The stickyEndpointID pointer is modified in place when the endpoint is not sticky.
+func (p *EndpointPool) FindStickyEndpoint(logger *slog.Logger, stickyEndpointID *string, mustBeSticky bool) *Endpoint {
+	var e *endpointElem
+	if *stickyEndpointID != "" {
```
maxmoehl: Can we turn this around? From what I can see we always return `nil` when the ID is empty and we also have no check against this not being nil which risks nil pointer panics. I'd prefer:


```suggestion
	if stickyEndpointID == nil || *stickyEndpointID == "" {
		return nil
	}
	// rest of the code
```

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -464,6 +465,48 @@ func (p *EndpointPool) findById(id string) *endpointElem {
...
+				}
+				return nil
+			}
+			e = nil
+		}
+
+		if e == nil && mustBeSticky {
+			if logger.Enabled(context.Background(), slog.LevelDebug) {
+				logger.Debug("endpoint-missing-but-request-must-be-sticky", slog.String("requested-endpoint", *stickyEndpointID))
+			}
```
maxmoehl: The explicit check for the log level only make sense if we plan on performing expensive preparation for the log (like in line 480), here we can just rely on the level check from slog:


```suggestion
			logger.Debug("endpoint-missing-but-request-must-be-sticky", slog.String("requested-endpoint", *stickyEndpointID))
```

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -464,6 +465,48 @@ func (p *EndpointPool) findById(id string) *endpointElem {
...
+			return nil
+		}
+
+		if !mustBeSticky {
+			if logger.Enabled(context.Background(), slog.LevelDebug) {
+				logger.Debug("endpoint-missing-choosing-alternate", slog.String("requested-endpoint", *stickyEndpointID))
+			}
```
maxmoehl:
```suggestion
			logger.Debug("endpoint-missing-choosing-alternate", slog.String("requested-endpoint", *stickyEndpointID))
```

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -464,6 +465,48 @@ func (p *EndpointPool) findById(id string) *endpointElem {
 	return p.index[id]
 }
 
+// FindStickyEndpoint attempts to find and return a sticky session endpoint.
+// If the endpoint is found and not overloaded, it returns the endpoint.
+// If mustBeSticky is true and the endpoint is missing or overloaded, it returns nil.
+// If mustBeSticky is false and the endpoint is missing or overloaded, it clears the stickyEndpointID and returns nil.
```
maxmoehl: Not sure I like this comment which essentially just explains the internals of the function.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -464,6 +465,48 @@ func (p *EndpointPool) findById(id string) *endpointElem {
...
+				logger.Debug("endpoint-missing-but-request-must-be-sticky", slog.String("requested-endpoint", *stickyEndpointID))
+			}
+			return nil
+		}
+
+		if !mustBeSticky {
+			if logger.Enabled(context.Background(), slog.LevelDebug) {
+				logger.Debug("endpoint-missing-choosing-alternate", slog.String("requested-endpoint", *stickyEndpointID))
+			}
+			*stickyEndpointID = ""
```
maxmoehl: I don't like this side-effect. I'd rather prefer to add logic that clears the value if the returned endpoint is nil at the call site. E.g. something like this:


```go
	e := r.pool.FindStickyEndpoint(r.logger, r.initialEndpoint, r.mustBeSticky)
	if e != nil {
		r.lastEndpoint = e
		return e
	}

	// clear initial endpoint as it is not usable
	r.initialEndpoint = ""

	if r.mustBeSticky {
		return nil
	}

	// ...
```

maxmoehl (commented):
Good catch! I do have some comments on the exact implementation, I get that most of it probably comes from the previous version but I'd like if we can use this opportunity to clean it up a little.

src/code.cloudfoundry.org/gorouter/route/pool.go
b1tamara: You are right. I rewrote the function description.

src/code.cloudfoundry.org/gorouter/route/pool.go
b1tamara: Refactored

src/code.cloudfoundry.org/gorouter/route/pool.go
b1tamara: Applied

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -466,45 +465,37 @@ func (p *EndpointPool) findById(id string) *endpointElem {
...
-			if mustBeSticky {
-				if logger.Enabled(context.Background(), slog.LevelDebug) {
-					logger.Debug("endpoint-overloaded-but-request-must-be-sticky", e.endpoint.ToLogData()...)
-				}
-				return nil
-			}
-			e = nil
-		}
+// The function returns nil if the sticky session endpoint is not found, or overloaded.
+func (p *EndpointPool) FindStickyEndpoint(logger *slog.Logger, stickyEndpointID string, mustBeSticky bool) *Endpoint {
```
maxmoehl, line 469: This looks much better now! And it's now somewhat obvious that this is just a verbose wrapper around `findById`. IIRC we said that logs are part of the contract so we probably can't easily refactor the remaining parts to collapse this into a single simple if, after all this function does only care about whether the endpoint must be sticky to emit some special logs.

maxmoehl: We agreed to stick to the current implementation to not cause any regressions.

maxmoehl (approved):

PR #534, opened 2026-03-04
Fix Cookie Attributes on Sticky Session Refresh

### LINKED ISSUE #529: Fix Session Affinity With Partitioned Session Cookies

hoffmaen:

Summary
---------------
When an application sets a partitioned `JSESSIONID` cookie, Gorouter sets a matching partitioned `__VCAP_ID__` cookie. However, when a subsequent request references a stale application instance and the new instance does not respond with a new `JSESSIONID` cookie, Gorouter had no way of knowing whether the client's existing `__VCAP_ID__` was partitioned — since cookie attributes are never transmitted from the client to the server. As a result, Gorouter always created a new unpartitioned `__VCAP_ID__` cookie in this case, leaving the client with both an orphaned partitioned and a new unpartitioned `__VCAP_ID__` cookie. This broke sticky session routing for partitioned cookies, as browsers such as Chrome would prefer the stale partitioned cookie over the new unpartitioned one.

This Pull Request introduces a new metadata cookie, `__VCAP_ID_META__`, which Gorouter sets alongside `__VCAP_ID__` whenever it creates a sticky session (i.e. the application responds with a matching session cookie such as `JSESSIONID`). The metadata cookie encodes all relevant `__VCAP_ID__` attributes — `Partitioned`, `Secure`, `SameSite`, and `Expiry` — as a compact `&`-separated `key=value` string (e.g. `expiry=1748001600&secure=1&samesite=none&partitioned=1`). In the stale instance scenario, when the new instance does not set a new `JSESSIONID` cookie, Gorouter reads `__VCAP_ID_META__` from the incoming request and reconstructs the refreshed `__VCAP_ID__` with the original attributes preserved.

The expiry lifetime is stored as an **absolute Unix timestamp** rather than a relative Max-Age value. This ensures that a stale instance refresh does not silently extend the user's session: the refreshed `__VCAP_ID__` receives an Expires header set to the original absolute point in time, so the remaining session lifetime is preserved exactly.

`__VCAP_ID_META__` is issued with the same attributes (`Secure`, `SameSite`, `Partitioned`, `Expiry`) as `__VCAP_ID__` itself, so the two cookies co-expire and share the same security constraints.

Closes #529 (Problem 1).

Backward Compatibility
---------------
Breaking Change? **No**
All clients that use sticky sessions will now additionally receive the `__VCAP_ID_META__` cookie alongside `__VCAP_ID__`. This is a net improvement: without it, sticky session routing for partitioned cookies in the stale instance scenario was already broken, and attributes such as Secure and SameSite were not preserved on refresh for any client. Clients that do not use sticky sessions are entirely unaffected. 

AI Usage
---------------
The work was done with the assistance of Claude (claude-sonnet-4-6, Anthropic) via claude code.


src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,76 +478,225 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+	secure         bool
+	sameSite       http.SameSite
+	absoluteExpiry int64 // Unix seconds; 0 = session cookie, -1 = delete (MaxAge < 0)
+	partitioned    bool
+}
+
+// encodeVcapMeta serialises a vcapMeta to the __VCAP_ID_META__ cookie value string.
+// The format is "&"-separated "key=value" pairs. Only non-default fields are included.
+// The "expiry" key is always present so the cookie is distinguishable from "no metadata".
+func encodeVcapMeta(m vcapMeta) string {
```
maxmoehl: You are already using `url.ParseQuery`, please use [`url.Values`](https://pkg.go.dev/net/url#Values) for encoding as well.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,76 +478,225 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+	case http.SameSiteStrictMode:
+		parts = append(parts, "samesite=strict")
+	}
+	return strings.Join(parts, "&")
+}
+
+// parseVcapMeta deserialises the __VCAP_ID_META__ cookie value.
+// Returns (meta, true) when the value contains a valid "expiry" key; otherwise (zero, false).
+func parseVcapMeta(value string) (vcapMeta, bool) {
+	params, err := url.ParseQuery(strings.ReplaceAll(value, "&", "&"))
```
maxmoehl: What's with the `strings.ReplaceAll`?

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,76 +478,225 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+			// Stale instance case: reconstruct attributes from __VCAP_ID_META__.
+			if m, found := findVcapMeta(requestCookies); found {
+				secure = m.secure
+				sameSite = m.sameSite
+				partitioned = m.partitioned
+				if m.absoluteExpiry > 0 {
+					expiry = time.Unix(m.absoluteExpiry, 0)
+				} else if m.absoluteExpiry < 0 {
+					maxAge = -1
+				}
```
maxmoehl: Hm, now we start messing with time, depending on the clients clock Expires might create weird situations in which MaxAge would have worked fine. I'd prefer either to always use MaxAge or copy what the original cookie did and deducting accordingly.

maxmoehl (dismissed):
I have some concerns when it comes to adding a second cookie:
1. Yet another cookie which has to be correlated, can be duplicated, etc. Based on your description the meta cookie encodes its own attributes, that would at least allow gorouter to clear it. But if you suddenly have two of those but one regular VCAP cookie it gets weird again.[^1]
2. This is a bit more vague but there might be environments in which cookies are tracked to some extent and a new cookie could cause confusion / issues.

I'd prefer if we could evolve the format of the `__VCAP_ID__` cookie instead:
* Add a feature flag for `vcap_id_cookie_v2` which takes the values `disabled` (default), `accept`, and `emit`.
* On `disabled` nothing changes, gorouter behaves as before.
* On `accept` gorouter tries to detect the new format and parse it, falling back to the current format if that fails.
* On `emit` it starts setting the VCAP_ID cookie according to the new format.
* Add `version=2` and `value=$value` to the cookie value.

All names are for illustration purposes only.

[^1]: Yes, I know, you can then just clear both. It still adds complexity.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
hoffmaen: I agree - we could expect client to be properly synced, and Expires is always UTC - however it's not necessary to introduce such logic at all. 
I refactored the code - we store and restore the two attributes independently now, keeping the semantics and logic with the client.

hoffmaen:
> I have some concerns when it comes to adding a second cookie:
> 
>     1. Yet another cookie which has to be correlated, can be duplicated, etc. Based on your description the meta cookie encodes its own attributes, that would at least allow gorouter to clear it. But if you suddenly have two of those but one regular VCAP cookie it gets weird again.[1](#user-content-fn-1-7ae6470452031de3247e0828ab074f13)
> 
>     2. This is a bit more vague but there might be environments in which cookies are tracked to some extent and a new cookie could cause confusion / issues.
> 
> 
> I'd prefer if we could evolve the format of the `__VCAP_ID__` cookie instead:
> 
>     * Add a feature flag for `vcap_id_cookie_v2` which takes the values `disabled` (default), `accept`, and `emit`.
> 
>     * On `disabled` nothing changes, gorouter behaves as before.
> 
>     * On `accept` gorouter tries to detect the new format and parse it, falling back to the current format if that fails.
> 
>     * On `emit` it starts setting the VCAP_ID cookie according to the new format.
> 
>     * Add `version=2` and `value=$value` to the cookie value.
> 
> 
> All names are for illustration purposes only.
> ## Footnotes
> 
>     1. Yes, I know, you can then just clear both. It still adds complexity. [↩](#user-content-fnref-1-7ae6470452031de3247e0828ab074f13)

I appreciate the idea, but I see a few drawbacks here:
1. Introducing a feature-flag, and then releasing the features in two phases means an additional delay to the rollout of the fix.
2. We do not know if anyone relies on the format of the `__VCAP_ID__` cookie as it is - it only stores the App Instance GUID. Changing the format is a breaking change.
3. I don't like introducing a feature flag for a thing that is not a feature - to me, this is a bug fix.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
 	endpoint *route.Endpoint,
 	originalEndpointId string,
 	secureCookies bool,
 	path string,
 	stickySessionCookieNames config.StringSet,
 	authNegotiateSticky bool,
+	logger *slog.Logger,
 ) {
+	sessionCookie, vcapCookie := getSessionCookiesFromResponse(response, stickySessionCookieNames)
+	if vcapCookie != nil {
```
a18e, line 525: A comment to explain the early return like "Sticky Session already established previously, nothing more to do" would be nice.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
 	request.Header.Set(router_http.CfInstanceIdHeader, value)
 }
 
+// stickySessionCase identifies which of the three mutually exclusive sticky-session paths applies to the current request/response pair.
+type stickySessionCase int
```
a18e: How about
```suggestion
type stickySessionScenario int
```

Also: With the new simplified structure I'm not sure if we need it at all...

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+const (
+	// stickySessionAuthNegotiate: first request of a Kerberos/SPNEGO handshake; fixed cookie defaults apply.
+	stickySessionAuthNegotiate stickySessionCase = iota
+	// sessionCookieInAppResponse: application set a JSESSIONID in the response; attributes are copied from it.
+	sessionCookieInAppResponse
```
a18e: Maybe this (or sth similar) makes the scenario more clear?

```suggestion
	newStickySessionRequested
```

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+		vcapCookie.SameSite = http.SameSiteStrictMode
+	case sessionCookieInAppResponse:
+		// Take cookie attributes from app session cookie in response
+		vcapCookie.Secure = sessionCookie.Secure
+		vcapCookie.SameSite = sessionCookie.SameSite
+		vcapCookie.Partitioned = sessionCookie.Partitioned
+		vcapCookie.MaxAge = sessionCookie.MaxAge
+		vcapCookie.Expires = sessionCookie.Expires
+		createMetaCookie = true
+	default: // case stickySessionStaleInstance
```
a18e: I know why we did this (to avoid the "stickySessionStaleInstance is always true" thing right?).
But a bit more information for others might be helpful.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+		// No sticky session to establish
+		return
+	}
 
-	secure := false
-	maxAge := 0
-	sameSite := http.SameSite(0)
-	expiry := time.Time{}
-	partitioned := false
+	// Create sticky session with default cookie attributes
```
a18e:
```suggestion
	// Create VCAP_ID cookie with default cookie attributes
```

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+		vcapCookie.Secure = true
+	}
 
-				maxAge = v.MaxAge
-				secure = v.Secure
-				sameSite = v.SameSite
-				expiry = v.Expires
-				partitioned = v.Partitioned
+	addCookieToResponse(response, vcapCookie)
+	if createMetaCookie {
```
a18e, line 580: Hey, since we're only setting this in the `case sessionCookieInAppResponse`, why not set it there directly and save ourselves the `var createMetaCookie`?

_Edit_: Because we would have to specify `if secureCookies {vcapCookie.Secure = true}` twice
_Edit2_: But we could just doo `....Secure = vcapCookie.Secure || secureCookies `

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+		addCookieToResponse(response, createVcapMetaCookie(vcapCookie))
+	}
+}
 
-				break
+// getSessionCookiesFromResponse returns the session cookie and __VCAP_ID__ cookie, when they exist in the response
+func getSessionCookiesFromResponse(response *http.Response, stickySessionCookieNames config.StringSet) (sessionCookie *http.Cookie , vcapIdCookie *http.Cookie) {
```
a18e:
```suggestion
func getSessionCookies(response *http.Response, stickySessionCookieNames config.StringSet) (sessionCookie *http.Cookie , vcapIdCookie *http.Cookie) {
```

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+	return nil
+}
 
-	if shouldSetVCAPID {
-		// right now secure attribute would as equal to the JSESSION ID cookie (if present),
-		// but override if set to true in config
-		if secureCookies {
-			secure = true
+// parseVcapMeta deserializes the __VCAP_ID_META__ cookie value.
+func parseVcapMeta(value string) (*vcapAttributes, error) {
```
a18e, line 617: Could we use some other kind of serialization that doesn't involve manual rules?

>   Option A: Intermediate JSON struct — keep vcapAttributes as-is, marshal/unmarshal via a separate vcapAttributesJSON with string fields and JSON tags. Simple, no custom marshaler needed.                                                                                                                                         
>   Option B: JSON tags + custom marshaler on vcapAttributes — add MarshalJSON/UnmarshalJSON to vcapAttributes. More self-contained but more boilerplate.      

Would probably be less performant though

> binary serialization (e.g. encoding/gob, protobuf, msgpack) would be faster and more compact than JSON, but you'd still be slower than the current url.Values approach for a payload this small.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+	}
 
-	if responseContainsAuthNegotiateHeader && authNegotiateSticky {
-		maxAge = AuthNegotiateHeaderCookieMaxAgeInSeconds
-		sameSite = http.SameSiteStrictMode
-	} else {
-		for _, v := range response.Cookies() {
-			if _, ok := stickySessionCookieNames[v.Name]; ok {
-				shouldSetVCAPID = true
+	if secureCookies {
```
a18e: Comment: point to global config value and that it overrides the per-cookie prefernce

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -476,75 +478,245 @@ func setRequestXCfInstanceId(request *http.Request, endpoint *route.Endpoint) {
...
+	for _, c := range cookies {
+		if c.Name == VcapMetaCookieId {
+			vcapMetaAttributes, err := parseVcapMeta(c.Value)
+			if err != nil {
+				logger.Error("vcap-id-meta-cookie-parse-error", log.ErrAttr(err))
+			}
+			return vcapMetaAttributes
 		}
 	}
+	logger.Info("vcap-id-meta-cookie-not-found")
```
a18e: Add comment that we expect to see these logs during the migration to the new cookie handling, but not after?

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
a18e: As discussed: We'll leave it as is, but will consider adding valueless params for booleans (e.g. `if params.Get("secure") { metaAttributesFromCookie.secure = true } `

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
hoffmaen: Setting attributes in one place, and adding cookies in another keeps the code structure clean. I think this is a legitimate reason to keep the few extra lines.

a18e (dismissed):
LGTM after several internal rounds of feedback/review/adjustments.
Please remember to squash and ideally add an AI usage disclaimer ([discussion pending](https://cloudfoundry.slack.com/archives/C02HNDJB31R/p1774426411842699))

a18e (approved):
Re-approving after test refactor

PR #535, opened 2026-03-05
RFC0055 Identity-Aware Routing

rkoster:
## RFC0055: Identity-Aware mTLS Routing

Implements Phase 1 (1a + 1b) of [RFC0055: App-to-App mTLS Routing](https://github.com/cloudfoundry/community/pull/1438).

**Tracking:** https://github.com/cloudfoundry/community/issues/1481
**Acceptance Testing Guide:** https://gist.github.com/rkoster/5b252b0edca606f10be2dbdcb81a796f

### What This Does

Enables GoRouter to enforce mutual TLS and identity-based authorization on a per-domain basis. Apps calling routes on configured mTLS domains must present a valid Diego instance identity certificate. GoRouter extracts the caller's app/space/org identity and checks it against route policies before forwarding the request.

### Phase 1a: mTLS Infrastructure

- Per-domain TLS configuration via `GetConfigForClient` callback (SNI-based)
- Domain-specific client certificate validation against configurable CA
- Domain-aware XFCC header handling with two formats:
  - `raw`: base64-encoded full certificate (~1.5KB)
  - `envoy`: compact `Hash=...;Subject="..."` format (~250 bytes)
- SNI/Host mismatch protection (prevents connection reuse attacks across domains)
- BOSH job properties for `router.domains`

### Phase 1b: Authorization

- Identity extraction from Diego instance identity certificates (Subject DN OUs + SPIFFE URIs)
- Pre-selection auth: validates mTLS domain, client cert presence, identity extraction
- Post-selection auth: enforces route policies (`scope` and `allowed_sources`) against selected endpoint
- Supports authorization at app, space, and org granularity
- Default deny when no route policies are configured
- RTR access logs emitted for denied requests (401/403)

### Key Design Decisions

- **Two-layer authorization**: Pre-selection (before endpoint is chosen) handles domain/cert/identity checks. Post-selection (after load balancer picks a backend) handles scope and route-policy checks against the specific endpoint's tags.
- **Feature is dormant by default**: No behavior change unless `router.domains` is configured in the BOSH manifest and a shared domain with `--enforce-access-rules` is created.
- **No regression on existing traffic**: Non-mTLS domains are completely unaffected.

### Testing

- Unit tests for all new handlers and config validation
- Integration tests for end-to-end mTLS routing flows
- BOSH template tests for configuration rendering
- CI runs `go fmt`, `go vet`, `staticcheck`, ginkgo with `--race`

### Configuration Example

```yaml
# BOSH manifest (via ops-file)
router:
  domains:
    - name: "*.apps.identity"
      ca_certs: "((diego_instance_identity_ca.certificate))"
      forwarded_client_cert: sanitize_set
      xfcc_format: envoy
```

### Related PRs

| Component | PR | Status |
|-----------|-----|--------|
| cloud_controller_ng | https://github.com/cloudfoundry/cloud_controller_ng/pull/4910 | Open |
| capi-release | https://github.com/cloudfoundry/capi-release/pull/625 | Open |
| CLI | https://github.com/cloudfoundry/cli/pull/3758 | Draft |

### Merge Ordering

All PRs are independently safe to merge — the feature is dormant without the ops-file and domain creation. No strict ordering required. Recommend merging around the same time once all are approved.

### AI Disclosure

This PR was developed with AI assistance. All code has been read and verified manually. Each error path, branch, and edge case has corresponding test coverage.


rkoster:
## Latest Update: RFC-Compliant Post-Selection Authorization

Implemented **breaking change** to replace pre-selection authorization with strict post-selection enforcement per RFC lines 475-517.

### Key Changes (commit cbf0695)

**Architecture:**
- ✅ Composable `PostSelectionHandler` interface for middleware pipeline
- ✅ Separation of pre-selection checks (SNI, route lookup, identity) from post-selection authorization
- ✅ Immediate 403 on authorization failure (non-retriable, per RFC)
- ✅ Post-selection scope checking with `:post-selection` suffix in metrics

**Implementation:**
- `handlers/post_selection_pipeline.go` - Infrastructure for composable checks
- `handlers/mtls_scope_auth.go` - Org/space boundary enforcement
- `handlers/mtls_access_rules_auth.go` - Access rules evaluation (cf:app:, cf:space:, etc.)
- `handlers/mtls_pre_auth.go` - Pre-selection checks only
- `handlers/mtls_auth_error.go` - Custom error type with Rule/Reason/HTTPStatus

**Test Coverage:**
- +44 new tests (14 scope + 17 access rules + 13 pipeline)
- +4 integration tests for shared route scenarios
- All 393 tests passing

### RFC Compliance

✅ **Intermittent 403s** - Expected for shared routes across scope boundaries (RFC-compliant)  
✅ **Error messages** - Include "caller org X does not match selected backend org Y"  
✅ **Strict enforcement** - Prevents unauthorized cross-scope access  

### Breaking Change

⚠️ This replaces the permissive pre-selection authorization entirely. No feature flag provided as this is a security improvement required by the RFC.

**Deprecated:**
- `handlers/mtls_authorization.go` (old implementation with migration notes)
- `route/pool.go` EndpointOrgIDs/SpaceIDs methods

### Integration Test Results

All integration tests compile successfully. Shared route scenarios validate:
- Intermittent 403s with `scope=space` (different spaces in same org)
- Always succeed with `scope=org` (same org, different spaces)
- Always fail with `scope=org` (different orgs)
- Per-endpoint access rules with intermittent behavior

Ready for full integration test run and review.

rkoster:
## Refactoring: AuthError for Future Extensibility

**Commit**: 4ff64b9d2

Renamed `MtlsAuthError` to `AuthError` to prepare for future authentication methods beyond mTLS, such as SPIFFE JWT tokens.

### Changes
- ✅ Renamed `handlers/mtls_auth_error.go` → `handlers/auth_error.go`
- ✅ Updated struct, constructor functions, and all references
- ✅ Changed error messages from "mTLS authorization denied" to "authorization denied"
- ✅ Updated all test files

### Benefits
- 🔮 **Future-proof**: Ready for SPIFFE JWT token authentication
- 🏗️ **Generic design**: Error type not tied to specific auth mechanism
- 🧩 **Reusable**: Can be used across different authentication methods
- ✨ **Clean**: Better naming convention for authorization errors

No functional changes - pure refactoring for extensibility.

rkoster:
Now that the RFC: https://github.com/cloudfoundry/community/pull/1438 has been merged let's start the review process.

ameowlia:
I was able to create a route that is both internal AND and mtls domain.

```
go run . create-shared-domain apps.identity.meow --enforce-route-policies --scope any --internal
Creating shared domain apps.identity.meow as admin...
OK

TIP: Domain 'apps.identity.meow' is a shared identity-aware domain with route policy enforcement enabled. Routes on this domain require route policies to allow traffic.
```

Is this intentional? I think we should limit it to only allowing one or the other. Same with the router group flag.

rkoster:
> I was able to create a route that is both internal AND and mtls domain.
> 
> ```
> go run . create-shared-domain apps.identity.meow --enforce-route-policies --scope any --internal
> Creating shared domain apps.identity.meow as admin...
> OK
> 
> TIP: Domain 'apps.identity.meow' is a shared identity-aware domain with route policy enforcement enabled. Routes on this domain require route policies to allow traffic.
> ```
> 
> Is this intentional? I think we should limit it to only allowing one or the other. Same with the router group flag.

Good point, could you leave this as feedback on: https://github.com/cloudfoundry/cloud_controller_ng/pull/4910 (since this constraint should be implemented there.

ameowlia:
So far I have just done black box testing. I am going to leave each thought as a different comment. 

~Issue: Removing policies does not work.~

~Here is what I did...~

~1. Push frontend and backend~
~2. Add identity route for backend~
~3. Frontend can’t talk to backend~
~4. Create route policy~
~5. Frontend CAN talk to backend~
~6. Remove policy~
~7. 🔥frontend CAN still talk to backend~
~8. 🔥I scaled frontend, got into the other instance, and that frontend CAN still talk to the backend too.~
~9. 🔥I ssh'd onto the backend app, and THAT app was able to talk to backend (despite there never being a route policy for that ever)~
~10. 🔥I pushed a new app, sshed onto that app, and that app was able to talk to the backend.~
~11. 🔥 I created a new identity route for the new app and other apps could automatically talk to it.~

~I tried multiple things to try to get a removed policy to stop working. I restarted apps, which did nothing. I monit restarted gorouter (my env only has one router), this fixed the issue once, but the second time I tested out this issue monit restarting even multiple times did not fix the issue.~


Update: Ruben and I paired on this. I was using too old of a diego version.

ameowlia:
Issue: I was surprised that there was nothing in the routes table to indicate that this is a special type of route. Or even to indicate which policies exist.

Currently the route entry looks like this...

```
  "backend.apps.identity": [
    {
      "address": "88.0.0.14:61066",
      "availability_zone": "az-0",
      "protocol": "http1",
      "tls": true,
      "ttl": 120,
      "tags": {
        "app_id": "8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0",
        "app_name": "backend",
        "component": "route-emitter",
        "instance_id": "0",
        "organization_id": "9ebf2ef5-5298-4faa-a554-c3236dfd3c11",
        "organization_name": "o",
        "process_id": "8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0",
        "process_instance_id": "3b94037e-d651-4705-6971-2fcd",
        "process_type": "web",
        "source_id": "8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0",
        "space_id": "10785795-fdf0-4178-9f74-d56276b8dfa4",
        "space_name": "s"
      },
      "private_instance_id": "3b94037e-d651-4705-6971-2fcd",
      "server_cert_domain_san": "3b94037e-d651-4705-6971-2fcd",
      "load_balancing_algorithm": "round-robin"
    }
  ],

```

I would expect the route table to reflect that this is a special kind of route.

rkoster:
Short update for the people following this PR. After debugging Amelia's environment a bit together, the current working theory is an older version of Diego, without generic route options support is causing the observed behavior.

ameowlia:
Things work much better with an up-to-date diego :)

ameowlia:
Issue: Inconsistent Access Log Fields

The new access log fields are conditionally present instead of always appearing with "-" when empty. Compare these logs:

```
backend.apps.identity - [2026-05-22T20:26:47.534183015Z] "GET / HTTP/2.0" 403 0 9 "-" "curl/7.81.0" "88.0.0.14:49748" "88.0.0.14:61030" x_forwarded_for:"88.0.0.14" x_forwarded_proto:"https" vcap_request_id:"bdc5651a-ba96-4ef0-56c5-094c4dee5ac8" response_time:0.000861 gorouter_time:0.000461 app_id:"8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0" app_index:"0" instance_id:"e677b14a-d066-4040-478b-5a39" x_cf_routererror:"-" tls_sni:"backend.apps.identity"
caller_app:"49c236fa-9b00-48b4-8b6f-f9d74c6a09e3"  <-- missing below
caller_space:"10785795-fdf0-4178-9f74-d56276b8dfa4" <-- missing below
caller_org:"9ebf2ef5-5298-4faa-a554-c3236dfd3c11" <-- missing below
auth:"denied" auth_rule:"route:no_route_policies"  <-- missing below
auth_denied_reason:"route has no route policies configured"   <-- missing below
x_b3_traceid:"bdc5651aba964ef056c5094c4dee5ac8" x_b3_spanid:"56c5094c4dee5ac8" x_b3_parentspanid:"-" b3:"bdc5651aba964ef056c5094c4dee5ac8-56c5094c4dee5ac8"

log-cache.sys.pcf.tasonvsphere.com - [2026-05-22T20:26:47.790646205Z] "GET /api/v1/read/8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0?envelope_types=LOG&start_time=1779481473833525000 HTTP/1.1" 200 0 2023 "-" "cf/8.8.0+c1b66a0.2024-08-21 (go1.22.5; arm64 darwin)" "192.168.115.1:60464" "88.0.0.31:8083" x_forwarded_for:"192.168.115.1" x_forwarded_proto:"https" vcap_request_id:"fa0351f9-2a13-4f7b-4e98-b120e449f99a" response_time:0.003497 gorouter_time:0.000298 app_id:"-" app_index:"-" instance_id:"090e5ff8-bf0f-444d-7c7e-bbf4bd30bb62" x_cf_routererror:"-" tls_sni:"log-cache.sys.pcf.tasonvsphere.com" x_b3_traceid:"fa0351f92a134f7b4e98b120e449f99a" x_b3_spanid:"4e98b120e449f99a" x_b3_parentspanid:"-" b3:"fa0351f92a134f7b4e98b120e449f99a-4e98b120e449f99a"
```

First log has no caller_* or auth* fields, second has them. Standard pattern is all fields should always present.

ameowlia:
Issue:  fields and names in access logs

here are the new fields

```
tls_sni:"backend.apps.identity"
caller_app:"49c236fa-9b00-48b4-8b6f-f9d74c6a09e3" 
caller_space:"10785795-fdf0-4178-9f74-d56276b8dfa4" 
caller_org:"9ebf2ef5-5298-4faa-a554-c3236dfd3c11" 
auth:"denied"
auth_rule:"route:no_route_policies"
auth_denied_reason:"route has no route policies configured"
```

I want to make it clear that these are only for this type of route. I suggest the following renames, but am open to suggestions.

caller_app --> caller_cf_app
caller_space --> caller_cf_space
caller_org --> caller_cf_org
auth_rule --> route_policy (the value for this should be empty when there are no route policies)

I don't know if these two values are needed and I would advocate for removing them
auth -->  Can't we tell that it is denied based on the status code?
auth_denied_reason -> isn't the reason always the same? either there are 0 route policies or there are 0 matching route policies?

ameowlia:
Issue: per request logs in gorouter.stdout.log

these logs are being logged per request in gorouter.stdout.log. This duplicates access log information and creates log volume amplification risk.
```
{
  "log_level": 1,
  "timestamp": "2026-05-22T20:36:14.559830443Z",
  "message": "mtls-route-policies-denied",
  "source": "vcap.gorouter",
  "data": {
    "route": "backend.apps.identity",
    "caller-app": "0fefb63d-8661-4305-95bd-c20efa21e1d7",
    "reason": "route-policies-deny",
    "endpoint": "88.0.0.14:61030"
  }
}

{
  "log_level": 1,
  "timestamp": "2026-05-22T20:36:14.560190128Z",
  "message": "post-selection-auth-denied",
  "source": "vcap.gorouter",
  "data": {
    "route-endpoint": {
      "ApplicationId": "8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0",
      "Addr": "88.0.0.14:61030",
      "Tags": {
        "app_id": "8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0",
        "app_name": "backend",
        "component": "route-emitter",
        "instance_id": "0",
        "organization_id": "9ebf2ef5-5298-4faa-a554-c3236dfd3c11",
        "organization_name": "o",
        "process_id": "8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0",
        "process_instance_id": "e677b14a-d066-4040-478b-5a39",
        "process_type": "web",
        "source_id": "8ba2b84a-7d48-45d6-a09b-0d9cd328c5e0",
        "space_id": "10785795-fdf0-4178-9f74-d56276b8dfa4",
        "space_name": "s"
      },
      "RouteServiceUrl": "",
      "AZ": "az-0"
    },
    "rule": "route:route_policies",
    "reason": "caller app 0fefb63d-8661-4305-95bd-c20efa21e1d7 not in route_policies",
    "endpoint": "88.0.0.14:61030"
  }
}
```

I suggest removing these logs (and any others that are per request).

rkoster:
> Issue: I was surprised that there was nothing in the routes table to indicate that this is a special type of route. Or even to indicate which policies exist.

Good point. This is addressed by commit [c058620](https://github.com/cloudfoundry/cloud_controller_ng/pull/4910/commits/c05862018cf489e94a92485a264df698732bc778) which added `?include=route_policies` support on the routes endpoints (`GET /v3/routes` and `GET /v3/routes/:guid`). The `IncludeRoutePoliciesDecorator` batch-loads policies via a single query and returns them in `included.route_policies`, following the same pattern as the existing `include=domain` and `include=space` decorators.

rkoster:
> Issue: Inconsistent Access Log Fields

> Issue: fields and names in access logs

Fixed — the identity-aware routing access log fields are now always emitted with "-" when empty (matching the standard pattern for all other log fields). The fields have been renamed: `caller_app` → `caller_cf_app`, `caller_space` → `caller_cf_space`, `caller_org` → `caller_cf_org`. The `auth`, `auth_rule`, and `auth_denied_reason` fields have been removed as suggested.

rkoster:
> Issue: per request logs in gorouter.stdout.log

Downgraded all 7 per-request denial log statements from INFO to DEBUG level (1c7f1c52). This eliminates the log volume amplification concern in production while keeping them available for operators who enable debug logging during troubleshooting.

The access log already captures denial information via `caller_cf_app`, `caller_cf_space`, `caller_cf_org` fields and HTTP status codes (421/403).

If you'd prefer these statements be removed entirely rather than downgraded, happy to do that instead — let me know.

ameowlia:
> The `auth`, `auth_rule`, and `auth_denied_reason` fields have been removed as suggested.

I didn't say auth_rule should be deleted! I like that one. 

I suggested a rename to route_policy. And when there is no matching routing policy I suggested using a "-" instead of "route:no_route_policies"

ameowlia:
> If you'd prefer these statements be removed entirely rather than downgraded, happy to do that instead — let me know.

I vote for removing them entirely since they duplicate the access log.

jobs/gorouter/spec
```diff
@@ -200,6 +200,31 @@ properties:
   router.only_trust_client_ca_certs:
     description: "When router.only_trust_client_ca_certs is true, router.client_ca_certs are the only trusted CA certs for client requests. When router.only_trust_client_ca_certs is false, router.client_ca_certs are trusted in addition to router.ca_certs and the CA certificates installed on the filesystem. This will have no affect if the `router.client_cert_validation` property is set to none."
     default: false
+  router.domains:
+    description: |
+      Array of domains requiring mutual TLS authentication. Each domain can have its own CA certificate pool, forwarded_client_cert mode, and xfcc_format.
+      For non-wildcard domains, the domain must match the request host exactly.
+      For wildcard domains (e.g., *.apps.identity), the wildcard must be the leftmost label and matches any single label.
```
ameowlia, line 209: Can you update the description to include info about how this interacts with the router.client_cert_validation property.

And update the router.client_cert_validation description to reference this property.

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -505,6 +505,57 @@ if p('router.client_ca_certs')
   params['client_ca_certs'] = client_ca_certs
 end
 
+if_p('router.domains') do |domains|
+  if !domains.is_a?(Array)
+    raise 'router.domains must be provided as an array'
+  end
```
ameowlia, line 511: I don't see a test for this.

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -505,6 +505,57 @@ if p('router.client_ca_certs')
...
+
+  processed_domains = []
+  domains.each do |domain_config|
+    if !domain_config.is_a?(Hash)
+      raise 'Each entry in router.domains must be a hash'
+    end
```
ameowlia, line 517: I don't see a test for this

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -505,6 +505,57 @@ if p('router.client_ca_certs')
...
+
+    if !domain_config.key?('name') || domain_config['name'].nil? || domain_config['name'].strip.empty?
+      raise 'Each entry in router.domains must have a "name" key'
```
ameowlia, line 520: test please

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -505,6 +505,57 @@ if p('router.client_ca_certs')
...
+    end
+
+    if !domain_config.key?('ca_certs') || domain_config['ca_certs'].nil? || domain_config['ca_certs'].strip.empty?
+      raise 'Each entry in router.domains must have a "ca_certs" key with certificate content'
```
ameowlia, line 524: test please

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -505,6 +505,57 @@ if p('router.client_ca_certs')
...
+  domains.each do |domain_config|
+    if !domain_config.is_a?(Hash)
+      raise 'Each entry in router.domains must be a hash'
+    end
+
+    if !domain_config.key?('name') || domain_config['name'].nil? || domain_config['name'].strip.empty?
+      raise 'Each entry in router.domains must have a "name" key'
+    end
+
+    if !domain_config.key?('ca_certs') || domain_config['ca_certs'].nil? || domain_config['ca_certs'].strip.empty?
```
ameowlia, line 523: test please

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -505,6 +505,57 @@ if p('router.client_ca_certs')
...
+    }
+
+    if domain_config.key?('forwarded_client_cert') && !domain_config['forwarded_client_cert'].nil?
+      valid_modes = ['always_forward', 'forward', 'sanitize_set']
+      mode = domain_config['forwarded_client_cert']
+      unless valid_modes.include?(mode)
+        raise "Invalid forwarded_client_cert mode '#{mode}' for domain '#{domain_config['name']}'. Must be one of: #{valid_modes.join(', ')}"
+      end
+      processed_entry['forwarded_client_cert'] = mode
+    end
```
ameowlia, line 539: test please

jobs/gorouter/templates/gorouter.yml.erb
```diff
@@ -505,6 +505,57 @@ if p('router.client_ca_certs')
...
+
+    if domain_config.key?('xfcc_format') && !domain_config['xfcc_format'].nil?
+      valid_formats = ['raw', 'envoy']
+      format = domain_config['xfcc_format']
+      unless valid_formats.include?(format)
+        raise "Invalid xfcc_format '#{format}' for domain '#{domain_config['name']}'. Must be one of: #{valid_formats.join(', ')}"
+      end
```
ameowlia, line 546: test please

src/code.cloudfoundry.org/gorouter/handlers/identity.go
```diff
@@ -0,0 +1,204 @@
...
+type CallerIdentity struct {
+	AppGUID   string
+	SpaceGUID string
+	OrgGUID   string
+}
+
+// identityHandler extracts the caller identity from the X-Forwarded-Client-Cert header
+// on mTLS domains. The identity is stored in the RequestInfo context for use by
+// authorization handlers.
+type identityHandler struct{}
```
ameowlia: Can we rename this CfIdentityHandler or something like that? To indicate that it is just for CF app certs?

src/code.cloudfoundry.org/gorouter/handlers/identity.go
```diff
@@ -0,0 +1,204 @@
...
+	// Extract identity from X-Forwarded-Client-Cert header
+	xfccHeader := r.Header.Get("X-Forwarded-Client-Cert")
+	if xfccHeader != "" {
+		identity, err := extractIdentityFromXFCC(xfccHeader)
+		if err == nil {
+			reqInfo.CallerIdentity = identity
+		}
+		// If extraction fails, continue without setting identity
+		// The authorization handler will deny access if identity is required
+	}
```
ameowlia, line 78: These values should only be extracted (1) when TLS was used (2) when using an mtls domain. Or else these could be spoofed.

src/code.cloudfoundry.org/gorouter/proxy/proxy.go
```diff
@@ -172,9 +185,12 @@ func NewProxy(
 		SkipSanitize(routeServiceHandler.(*handlers.RouteService)),
 		ForceDeleteXFCCHeader(routeServiceHandler.(*handlers.RouteService), cfg.ForwardedClientCert, logger),
 		cfg.ForwardedClientCert,
+		cfg,
 		logger,
 		errorWriter,
 	))
+	n.Use(handlers.NewIdentity())
+	n.Use(handlers.NewMtlsPreAuth(cfg, logger))
```
ameowlia: Could the NewMtlsPreAuth handler be moved about the NewClientCert handler? that way if the MtlsPreAuth returns a 421 we save on certificate processing time.

src/code.cloudfoundry.org/gorouter/handlers/identity.go
```diff
@@ -0,0 +1,204 @@
...
+			return nil, errors.New("malformed Cert field in XFCC header")
+		}
+		pemData := xfcc[certStart : certStart+certEnd]
+
+		// Decode PEM block
+		block, _ := pem.Decode([]byte(pemData))
+		if block == nil {
+			return nil, errors.New("failed to decode PEM certificate")
+		}
+		certDER = block.Bytes
```
ameowlia: Shouldn't the format be set in the bosh property? And if it is provided incorrectly then it should be rejected.

src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -579,6 +603,48 @@ func (p *EndpointPool) IsEmpty() bool {
...
+// All endpoints in a pool should have the same ApplicationId.
+func (p *EndpointPool) ApplicationId() string {
+	p.Lock()
+	defer p.Unlock()
+
+	if len(p.endpoints) == 0 {
+		return ""
+	}
+
+	return p.endpoints[0].endpoint.ApplicationId
```
ameowlia, line 654: There is a chance that routes are stale (by design) and these route policies are out of date.

I think it would be better to store route policies at the pool level instead of on the endpoint object. This would also reduce mutex contention by avoiding the need to acquire the pool lock twice per request.

jobs/gorouter/spec
rkoster: Updated both descriptions. `router.client_cert_validation` now notes that `router.domains` enforce their own mTLS requirement independently. `router.domains` description now clarifies that client certificates are always required for configured domains regardless of `router.client_cert_validation`.

jobs/gorouter/templates/gorouter.yml.erb
rkoster: Added tests in `spec/gorouter_templates_spec.rb` covering each validation branch: non-array input, non-hash entry, missing/empty name, missing/empty ca_certs, invalid forwarded_client_cert mode, and invalid xfcc_format value. All 15 domain-related specs passing.

jobs/gorouter/templates/gorouter.yml.erb
rkoster: Added tests in `spec/gorouter_templates_spec.rb` covering this validation branch. All 15 domain-related specs passing.

jobs/gorouter/templates/gorouter.yml.erb
rkoster: Added tests in `spec/gorouter_templates_spec.rb` covering this validation branch. All 15 domain-related specs passing.

jobs/gorouter/templates/gorouter.yml.erb
rkoster: Added tests in `spec/gorouter_templates_spec.rb` covering this validation branch. All 15 domain-related specs passing.

jobs/gorouter/templates/gorouter.yml.erb
rkoster: Added tests in `spec/gorouter_templates_spec.rb` covering this validation branch. All 15 domain-related specs passing.

jobs/gorouter/templates/gorouter.yml.erb
rkoster: Added tests in `spec/gorouter_templates_spec.rb` covering this validation branch. All 15 domain-related specs passing.

jobs/gorouter/templates/gorouter.yml.erb
rkoster: Added tests in `spec/gorouter_templates_spec.rb` covering this validation branch. All 15 domain-related specs passing.

src/code.cloudfoundry.org/gorouter/handlers/identity.go
rkoster: Renamed to `cfIdentityHandler` / `NewCfIdentity(cfg)` to make clear this handles CF app instance identity certificates specifically.

src/code.cloudfoundry.org/gorouter/handlers/identity.go
rkoster: Fixed in ac2e87e88 — the handler now checks `r.TLS != nil` and `config.GetMtlsDomainConfig(host) != nil` before extracting identity from the XFCC header. Requests on non-TLS connections or non-mTLS domains skip extraction entirely.

src/code.cloudfoundry.org/gorouter/proxy/proxy.go
rkoster: Done — moved `NewMtlsPreAuth` above `NewClientCert` in the handler chain. A 421 from MtlsPreAuth now short-circuits before any certificate processing happens.

src/code.cloudfoundry.org/gorouter/handlers/identity.go
rkoster: Addressed — `extractIdentityFromXFCC` now takes the configured `xfcc_format` from the domain config and dispatches to format-specific parsing (`extractIdentityFromEnvoyXFCC` or `extractIdentityFromRawXFCC`). Auto-detection has been removed. Invalid format values are already rejected at config validation time in `processMtlsDomains()`.

src/code.cloudfoundry.org/gorouter/handlers/mtls_pre_auth.go
```diff
@@ -0,0 +1,163 @@
...
+	// Exact match
+	if hostname == domainPattern {
+		return true
+	}
+	// Wildcard match - must match single label only
+	if strings.HasPrefix(domainPattern, "*.") {
+		suffix := domainPattern[1:] // Remove the '*', suffix = ".apps.identity"
+		if !strings.HasSuffix(hostname, suffix) {
+			return false
+		}
```
ameowlia: I think this is case sensitive, but URLs are case insensitive.

src/code.cloudfoundry.org/gorouter/route/pool.go
rkoster: Done in 26c34851 — moved `routePolicyScope` and `routePolicies` from endpoint to pool level. The pool now stores these as fields that get updated on Put/re-Put. Added unit tests covering initial state, updates, and persistence after endpoint removal.

This eliminates the stale data concern since all endpoints on a route share the same policies, and reduces lock contention by reading from pool fields instead of iterating endpoints.

src/code.cloudfoundry.org/gorouter/handlers/mtls_pre_auth.go
rkoster: Fixed in 82e9fa172 — domain matching is now case-insensitive per RFC 1035 (https://www.ietf.org/rfc/rfc1035.txt). Both `processMtlsDomains()` and `GetMtlsDomainConfig()` normalize hostnames to lowercase. Added unit tests covering uppercase, mixed case, and case with port combinations.

rkoster:
> I vote for removing them entirely since they duplicate the access log.

Done — removed the per-request denial log statements entirely in 26c34851. The access log already captures all denial information via the `caller_cf_*` fields, `route_policy` field, and HTTP status codes.

src/code.cloudfoundry.org/gorouter/handlers/mtls_sni_check.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// SNI/Host consistent with mTLS domain — continue to certificate processing
+	next(w, r)
+}
+
+// domainMatches checks if a hostname matches a domain pattern (supports wildcard domains).
+// Wildcard patterns (*.domain) only match a single DNS label, not multiple levels.
+func domainMatches(hostname, domainPattern string) bool {
+	if hostname == domainPattern {
```
ameowlia, line 86: I think there is still a case sensitivity issue here because there is nothing that lowercases the hostname.

src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -902,6 +929,59 @@ func (c *Config) buildClientCertPool() error {
...
+			return fmt.Errorf("domains[%d].ca_certs is required", i)
+		}
+
+		// Validate domain is not empty
+		if domain.Domain == "" {
+			return fmt.Errorf("domains[%d].domain is required", i)
+		}
+
+		// Store with lowercase key for case-insensitive matching (RFC 1035)
+		c.mtlsDomainMap[strings.ToLower(domain.Domain)] = domain
```
ameowlia: There is still a casing issue here. The map key is lowercased. But domain.Domain is not. 

I think it should be consistent and lowercased everywhere.

ameowlia:
I found more per-request logs, which goes against the logging standards for gorouter.

mtls_sni_check.go:52 - Non-mTLS domain with cert required
mtls_sni_check.go:69 - mTLS domain without proper cert validation
mtls_scope_auth.go:41 - Route pool missing during scope auth
mtls_route_policies_auth.go:72 - Route pool missing during policies auth
router.go:399 - mTLS domain detected during TLS handshake

All of these should already result in access logs (I think). Thus these logs are duplicative and should be deleted.

src/code.cloudfoundry.org/gorouter/proxy/proxy.go
```diff
@@ -168,13 +181,17 @@ func NewProxy(
...
 	n.Use(handlers.NewClientCert(
 		SkipSanitize(routeServiceHandler.(*handlers.RouteService)),
 		ForceDeleteXFCCHeader(routeServiceHandler.(*handlers.RouteService), cfg.ForwardedClientCert, logger),
 		cfg.ForwardedClientCert,
+		cfg,
 		logger,
 		errorWriter,
 	))
+	n.Use(handlers.NewCfIdentity(cfg))
+	n.Use(handlers.NewMtlsPreAuth(cfg, logger))
```
ameowlia, line 195: Can you add ifs so that these handlers aren't always run? I suggest something like...

shouldRunMtlsHandlers := len(config.Domains) > 0 || config.GlobalMtlsEnabled

ameowlia: actually DOES this code work when config.GlobalMtlsEnabled is true? if that is true, should all domains automatically work with route policies? Or do you want to force people to configure the domains in that case?

src/code.cloudfoundry.org/gorouter/handlers/mtls_pre_auth.go
rkoster: The earlier fix in 846c45b was incomplete. Fixed properly in d3f4135 — `domainMatches()` now lowercases both hostname and pattern before comparison, and `domain.Domain` is normalized to lowercase when stored (not just the map key). Added 9 tests covering mixed-case scenarios.

src/code.cloudfoundry.org/gorouter/handlers/mtls_sni_check.go
rkoster: Fixed in d3f4135 — `domainMatches()` now normalizes both `hostname` and `domainPattern` to lowercase before comparison, per RFC 1035. Added tests for uppercase hostnames, mixed-case suffixes, and mixed-case domain patterns.

src/code.cloudfoundry.org/gorouter/config/config.go
rkoster: Fixed in d3f4135 — `domain.Domain` is now normalized to lowercase when storing, not just the map key. This ensures downstream code (like `domainMatches`) receives consistent lowercase values. Added tests verifying `cfg.Domain` returns lowercase even when configured with mixed case.

rkoster:
> I found more per-request logs, which goes against the logging standards for gorouter.
>
> mtls_sni_check.go:52 - Non-mTLS domain with cert required
> mtls_sni_check.go:69 - mTLS domain without proper cert validation
> mtls_scope_auth.go:41 - Route pool missing during scope auth
> mtls_route_policies_auth.go:72 - Route pool missing during policies auth
> router.go:399 - mTLS domain detected during TLS handshake

Removed all 5 in f100947. Also found and removed one additional per-request log: `clientcert.go:63` (`using-mtls-domain-xfcc-config`).

src/code.cloudfoundry.org/gorouter/proxy/proxy.go
rkoster: Done, I went with the no-op handler pattern instead of adding conditionals in proxy.go.

When I tried to write tests for conditional wrapping in proxy.go, I ran into a code smell: the tests needed intimate knowledge of handler internals to verify observable behavior differences. That coupling suggested the abstraction was in the wrong place.

So instead, the handler constructors now return a no-op handler when `len(cfg.Domains) == 0`. This keeps the decision inside the handler package and is straightforward to test. See commit a4c165a7b.

Regarding `GlobalMtlsEnabled`: I think requiring explicit `router.domains` config makes sense for identity-aware routing. The feature is strongest when a dedicated CA is configured for the identity domain. With global `client_cert_validation: require`, operators would need to add off-platform client CAs to the trust store to allow external traffic — but then those external certs could potentially spoof CF identity OUs (`app:`, `space:`, `org:`), weakening the trust model. A dedicated domain with only the Diego CA keeps the identity guarantees tight.

src/code.cloudfoundry.org/gorouter/accesslog/schema/access_log_record.go
```diff
@@ -316,6 +327,23 @@ func (r *AccessLogRecord) makeRecord(performTruncate bool) []byte {
...
+	b.WriteString(` caller_cf_space:`)
+	b.WriteDashOrStringValue(r.CallerCFSpace)
+	// #nosec G104
+	b.WriteString(` caller_cf_org:`)
+	b.WriteDashOrStringValue(r.CallerCFOrg)
+	// #nosec G104
+	b.WriteString(` route_policy:`)
+	b.WriteDashOrStringValue(r.RoutePolicy)
+
 	r.addExtraHeaders(b, performTruncate)
```
ameowlia, line 345: I forgot that we have this idea of extra access log fields.

https://github.com/cloudfoundry/routing-release/blob/90b7cdb9d4ae539cb92879691b6493a297c419b4/jobs/gorouter/spec#L336-L343

What do you think about moving all of these new access log fields into that category? That way only foundations that use this feature will have these extra logs.

src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -902,6 +929,62 @@ func (c *Config) buildClientCertPool() error {
...
+	for i := range c.Domains {
+		domain := &c.Domains[i]
+		domain.RequireClientCert = true
+
+		// Validate forwarded_client_cert mode
+		if domain.ForwardedClientCert == "" {
+			domain.ForwardedClientCert = SANITIZE_SET // Default to most secure
+		}
+		if !slices.Contains(AllowedForwardedClientCertModes, domain.ForwardedClientCert) {
+			return fmt.Errorf("domains[%d].forwarded_client_cert must be one of %v",
```
ameowlia, line 945: missing test for this error

ameowlia: Also we should probably do some capitalization normalization in the gorouter spec so it is always lowercase by the time it gets here

src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -902,6 +929,62 @@ func (c *Config) buildClientCertPool() error {
...
+				i, AllowedForwardedClientCertModes)
+		}
+
+		// Validate xfcc_format
+		if domain.XFCCFormat == "" {
+			domain.XFCCFormat = XFCC_FORMAT_RAW // Default to raw for backwards compatibility
+		}
+		if !slices.Contains(AllowedXFCCFormats, domain.XFCCFormat) {
+			return fmt.Errorf("domains[%d].xfcc_format must be one of %v",
```
ameowlia, line 954: missing test for this error case

src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -902,6 +929,62 @@ func (c *Config) buildClientCertPool() error {
...
+		if domain.XFCCFormat != XFCC_FORMAT_RAW && domain.ForwardedClientCert == ALWAYS_FORWARD {
+			return fmt.Errorf("domains[%d].xfcc_format has no effect when forwarded_client_cert is 'always_forward'; remove xfcc_format or change forwarded_client_cert to 'sanitize_set'",
+				i)
+		}
+
+		// Build CA pool for this domain
+		if domain.CACerts != "" {
+			pool := x509.NewCertPool()
+			if !pool.AppendCertsFromPEM([]byte(domain.CACerts)) {
+				return fmt.Errorf("domains[%d].ca_certs contains invalid certificates", i)
```
ameowlia, line 966: missing test case for this error

src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -902,6 +929,62 @@ func (c *Config) buildClientCertPool() error {
...
+			}
+			domain.CAPool = pool
+		} else {
+			return fmt.Errorf("domains[%d].ca_certs is required", i)
```
ameowlia, line 970: missing test case for this error

src/code.cloudfoundry.org/gorouter/config/config.go
```diff
@@ -902,6 +929,62 @@ func (c *Config) buildClientCertPool() error {
...
+		}
+
+		// Validate domain is not empty
+		if domain.Domain == "" {
+			return fmt.Errorf("domains[%d].domain is required", i)
```
ameowlia, line 975: missing test case for this error

src/code.cloudfoundry.org/gorouter/config/config_test.go
```diff
@@ -2043,6 +2043,244 @@ drain_timeout: 60s
...
+			})
+
+			It("matches with uppercase host and port", func() {
+				cfg := config.GetMtlsDomainConfig("EXACT.EXAMPLE.COM:443")
+				Expect(cfg).ToNot(BeNil())
+				Expect(cfg.Domain).To(Equal("exact.example.com"))
+			})
+		})
+
+		Context("when domain is configured with mixed case (Thread 16: domain.Domain not lowercased)", func() {
```
ameowlia: lol maybe change the context here to remove "Thread 16"

src/code.cloudfoundry.org/gorouter/handlers/access_log.go
```diff
@@ -92,11 +83,18 @@ func (a *accessLog) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http
...
+	if reqInfo.CallerIdentity != nil {
+		alr.CallerCFApp = reqInfo.CallerIdentity.AppGUID
+		alr.CallerCFSpace = reqInfo.CallerIdentity.SpaceGUID
+		alr.CallerCFOrg = reqInfo.CallerIdentity.OrgGUID
+	}
+	if reqInfo.AuthResult != nil {
+		alr.RoutePolicy = reqInfo.AuthResult.Rule
 	}
+	alr.TlsSNI = reqInfo.TlsSNI
+
```
ameowlia, line 105: add tests for this handler

src/code.cloudfoundry.org/gorouter/accesslog/schema/access_log_record.go
rkoster: Done. Moved `tls_sni`, `caller_cf_app`, `caller_cf_space`, `caller_cf_org`, and `route_policy` into the `extra_access_log_fields` opt-in mechanism. They now follow the same pattern as `backend_time`, `dial_time`, etc. — only emitted when explicitly listed in the manifest. Updated the spec description to include all five new field names in the available fields list. Added three test cases covering populated values, empty (dash) values, and absence from output when not listed in `ExtraFields`.

src/code.cloudfoundry.org/gorouter/config/config.go
rkoster: Done on both counts. Added a test for the invalid `forwarded_client_cert` error (commit cbe790a). Also added `downcase` in the ERB template for both `forwarded_client_cert` and `xfcc_format` before validation, so mixed-case operator input is normalized before it reaches Go.

src/code.cloudfoundry.org/gorouter/config/config.go
rkoster: Added in cbe790a.

src/code.cloudfoundry.org/gorouter/config/config.go
rkoster: Added in cbe790a.

src/code.cloudfoundry.org/gorouter/config/config.go
rkoster: Added in cbe790a.

src/code.cloudfoundry.org/gorouter/config/config.go
rkoster: Added in cbe790a.

src/code.cloudfoundry.org/gorouter/config/config_test.go
rkoster: Removed in cbe790a.

src/code.cloudfoundry.org/gorouter/handlers/access_log.go
rkoster: Added in 441bf8da. Tests cover `CallerIdentity` → `CallerCFApp/Space/Org`, `AuthResult` → `RoutePolicy`, and `TlsSNI` propagation — including nil/empty cases for each.

rkoster: Added here: https://github.com/cloudfoundry/routing-release/pull/535/changes/ea8b9388274615ccaf888cca8a56e9b54c74abef

src/code.cloudfoundry.org/gorouter/handlers/identity_test.go
```diff
@@ -0,0 +1,393 @@
...
+			request = test_util.NewRequest("GET", "regular.example.com", "/", nil)
+			request.TLS = &tls.ConnectionState{}
+			cert := generateTestCert("app:should-not-extract")
+			request.Header.Set("X-Forwarded-Client-Cert", buildGoRouterXFCCHeader(cert))
+		})
+
+		It("skips identity extraction and calls next", func() {
+			handler.ServeHTTP(recorder, request, nextHandler)
+			Expect(nextCalled).To(BeTrue())
+		})
```
ameowlia: Can this test be improved? This only tests that it calls next. It doesnt test that the rest of the functionality doesn't run.

src/code.cloudfoundry.org/gorouter/handlers/identity_test.go
```diff
@@ -0,0 +1,393 @@
...
+	})
+
+	Context("when RequestInfo is not in context", func() {
+		It("calls next handler without setting identity", func() {
+			handler.ServeHTTP(recorder, request, nextHandler)
+
+			Expect(nextCalled).To(BeTrue())
+			Expect(recorder.Code).To(Equal(http.StatusOK))
+		})
+	})
```
ameowlia, line 134: Can this test be improved? This only tests that it calls next. It doesnt test that the identity is not set.

src/code.cloudfoundry.org/gorouter/handlers/identity_test.go
```diff
@@ -0,0 +1,393 @@
...
+		BeforeEach(func() {
+			request.TLS = nil
+			cert := generateTestCert("app:should-not-extract")
+			request.Header.Set("X-Forwarded-Client-Cert", buildGoRouterXFCCHeader(cert))
+		})
+
+		It("skips identity extraction and calls next", func() {
+			handler.ServeHTTP(recorder, request, nextHandler)
+			Expect(nextCalled).To(BeTrue())
+		})
```
ameowlia: Can this test be improved? This only tests that it calls next. It doesnt test that the rest of the functionality doesn't run.

src/code.cloudfoundry.org/gorouter/handlers/mtls_route_policies_auth_test.go
```diff
@@ -0,0 +1,577 @@
...
+				authErr, ok := err.(*handlers.AuthError)
+				Expect(ok).To(BeTrue())
+				Expect(authErr.Rule).To(Equal("route:route_policies"))
+			})
+		})
+
+		// ── Per-endpoint policies (shared routes with different backends) ──
+
+		Context("when endpoints on same route have different route policies", func() {
+			It("uses selected endpoint's policies, not pool-level policies", func() {
```
ameowlia: I think it should always use the pool level policies, which should always be the most up-to-date.

src/code.cloudfoundry.org/gorouter/handlers/postselection/mtls_route_policies_auth_test.go
```diff
@@ -0,0 +1,577 @@
...
+				reqInfo.CallerIdentity = &handlers.CallerIdentity{
+					AppGUID: "caller-app",
+				}
+
+				err := handler.Check(endpoint, reqInfo)
+				Expect(err).To(BeNil())
+				Expect(reqInfo.AuthResult.Rule).To(Equal("route:cf:any"))
+			})
+
+			It("skips malformed rules and evaluates valid ones", func() {
```
ameowlia, line 549: Hmmm. I think gorouter should log when there is a malformed rule. 

Or else it would be so difficult to determine that gorouter is not evaluating something properly.

src/code.cloudfoundry.org/gorouter/handlers/postselection/mtls_scope_auth.go
```diff
@@ -0,0 +1,107 @@
...
+// intermittent 403 errors are expected as the RFC acknowledges this as the
+// tradeoff for strict per-endpoint authorization.
+type MtlsScopeAuth struct {
+	config *config.Config
+	logger *slog.Logger
+}
+
+// NewMtlsScopeAuth creates a new post-selection scope authorization handler.
+// Returns NoopPostSelectionHandler when no mTLS domains are configured.
+func NewMtlsScopeAuth(cfg *config.Config, logger *slog.Logger) PostSelectionHandler {
```
ameowlia, line 28: I don't love that this and others (PostSelectionHandler, PostSelectionPipeline, AuthError, NoopPostSelectionHandler, possibly more that I missed) are being called handlers and are in the handler dir, when they aren't real HTTP handlers. 

What do you think of putting them in a new directory /handers/post_selection or something to keep them clearly separate from the HTTP handlers?

src/code.cloudfoundry.org/gorouter/handlers/identity_test.go
rkoster: Improved in 41d7c5178. The spec now runs the handler through a NewRequestInfo middleware chain and, after next is called, asserts requestInfo.CallerIdentity is nil. So it verifies the handler actually skipped identity extraction (no identity stored) on a non-TLS request carrying a valid XFCC header, not just that it called next.

src/code.cloudfoundry.org/gorouter/handlers/identity_test.go
rkoster: Improved in 41d7c5178. For a non-mTLS host carrying a valid XFCC header, the spec now asserts requestInfo.CallerIdentity is nil after next is called — confirming the handler leaves identity unset on domains outside the configured mTLS set, rather than only checking that next ran.

src/code.cloudfoundry.org/gorouter/handlers/identity_test.go
rkoster: Improved in 41d7c5178. With no NewRequestInfo middleware in front of the handler (so the context has no RequestInfo), the spec now asserts handlers.ContextRequestInfo(request) returns an error after next is called — proving the handler bailed out cleanly on a valid XFCC header without panicking and without creating a RequestInfo or storing any identity, not just that next ran.

src/code.cloudfoundry.org/gorouter/handlers/mtls_route_policies_auth_test.go
rkoster: Agreed — changed in 41d7c5178. MtlsRoutePoliciesAuth.Check() now reads policies from reqInfo.RoutePool.RoutePolicies() instead of the selected endpoint's per-endpoint copy, which can be stale when a route is shared across backends. The specs were rewritten to verify authorization uses the pool-level policies (including a case proving a stale per-endpoint copy no longer affects the decision).

src/code.cloudfoundry.org/gorouter/handlers/postselection/mtls_route_policies_auth_test.go
rkoster: Good call — and gorouter does log this. In `evaluateRoutePolicies`, the default (unrecognized-syntax) branch emits a `warn`-level `malformed-route-policy` log including the offending policy string, so a misconfigured rule is visible in the router log even though it's skipped during evaluation:

```go
default:
    // Unrecognized rule syntax — logged so operators can detect
    // a misconfigured route policy that's being silently skipped.
    h.logger.Warn("malformed-route-policy", slog.String("policy", policy))
```

This test asserts that behavior via `gbytes.Say("malformed-route-policy")`.

As we discussed on Slack: since Cloud Controller validates route-policy syntax at creation time ([`route_policy_create_message.rb`](https://github.com/cloudfoundry/cloud_controller_ng/blob/c05862018cf489e94a92485a264df698732bc778/app/messages/route_policy_create_message.rb#L5)), a malformed rule reaching gorouter should be rare, so the warn-level router log is sufficient and we don't need to surface it on the access-log line. Resolving accordingly.

src/code.cloudfoundry.org/gorouter/handlers/postselection/mtls_scope_auth.go
rkoster: Great suggestion — done. All the post-selection types moved out of the HTTP-handler package into a new subpackage `handlers/postselection`:

- `PostSelectionHandler` (interface) + `PostSelectionPipeline` — `post_selection_pipeline.go`
- `MtlsScopeAuth`, `MtlsRoutePoliciesAuth` — `mtls_scope_auth.go`, `mtls_route_policies_auth.go`
- `AuthError` (+ `NewAuthError`/`NewAuthErrorWithStatus`) — `auth_error.go`
- `NoopPostSelectionHandler` — `noop.go`

I went with `postselection` (no underscore) to follow Go's package-naming convention (lowercase, no underscores).

The request-scoped types `RequestInfo`, `CallerIdentity`, and `AuthResult` intentionally stay in `handlers` — they hang off the request context and are shared with the real HTTP handlers, so `postselection` imports `handlers` for those (no import cycle).

src/code.cloudfoundry.org/gorouter/integration/identity_aware_routing_test.go
```diff
@@ -0,0 +1,910 @@
...
+
+			It("requires a client certificate", func() {
+				// Register route on mTLS domain
+				testState.register(backendApp, mtlsDomain)
+
+				// Attempt request without client certificate
+				req, client := testState.newMtlsGetRequest(fmt.Sprintf("https://%s", mtlsDomain))
+				_, err := client.Do(req)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("tls"))
```
ameowlia: Is there a more specific error/string this can match on? Or does it change? This `Expect(err.Error()).To(ContainSubstring("tls"))` is used in multiple places in file. So if you change one, please change them all.

src/code.cloudfoundry.org/gorouter/integration/identity_aware_routing_test.go
```diff
@@ -0,0 +1,910 @@
...
+						resp, err := client.Do(req)
+						Expect(err).NotTo(HaveOccurred())
+						Expect(resp.StatusCode).To(Equal(http.StatusForbidden))
+						resp.Body.Close()
+					}
+				})
+			})
+
+			Context("when shared route has app-specific route policies", func() {
+				It("allows only the specified app and denies others (per-endpoint rules)", func() {
```
ameowlia: I think in the last review we got rid of per-endpoint rules and said it should only use the pool rules.

src/code.cloudfoundry.org/gorouter/integration/identity_aware_routing_test.go
rkoster: Fixed in 686f78ffb — both occurrences now use specific strings:

- line 88 (no client cert): `ContainSubstring("tls: certificate required")`
- line 157 (unknown CA): `ContainSubstring("tls: unknown certificate authority")`

These are the Go TLS alert names and have been stable since Go 1.15.

src/code.cloudfoundry.org/gorouter/integration/identity_aware_routing_test.go
rkoster: You're right, and there's a deeper issue here: the pool's `routePolicies` field is last-writer-wins — each `Add`/`Update` unconditionally overwrites it with the new endpoint's policies. In this test, after backend-2 registers, the pool holds only `["cf:app:allowed-app-2"]`, so a caller with `allowed-app-1` would always get 403. The test was asserting intermittent allow/deny behaviour that the current pool implementation makes impossible — it would always fail at `Expect(successCount).To(BeNumerically(">", 0))`.

Last-wins is intentional: in production, CAPI enforces policies per-route at registration time, so all backends of a route always carry identical policies — registrations are idempotent updates. The test was constructing an artificial scenario that can't happen in production.

Removed the test and added a comment in `pool.go` explaining this invariant in e3676647b.

This should have been caught by CI earlier — the `ready-to-run` label wasn't picked up by GitHub Actions, so the integration tests weren't triggered. Sorry for that.

ameowlia:
✅ After identity domain is created, but before route policies are created, it shows up in the gorouter routing table...

```
  "backend.apps.identity": [
    {
      "address": "10.10.0.10:61000",
      "availability_zone": "null",
      "protocol": "http1",
      "tls": true,
      "ttl": 120,
      "tags": {
        "app_id": "07a9f703-300b-40e3-81be-2f0060f7f676",
        "app_name": "backend",
        "component": "route-emitter",
        "instance_id": "0",
        "organization_id": "f77f7566-0895-42f9-b0a7-cde95989fb1b",
        "organization_name": "o",
        "process_id": "07a9f703-300b-40e3-81be-2f0060f7f676",
        "process_instance_id": "8124e938-9e7c-4409-7732-81a1",
        "process_type": "web",
        "source_id": "07a9f703-300b-40e3-81be-2f0060f7f676",
        "space_id": "bf561092-f646-408c-ab70-3e95f597f076",
        "space_name": "s"
      },
      "private_instance_id": "8124e938-9e7c-4409-7732-81a1",
      "server_cert_domain_san": "8124e938-9e7c-4409-7732-81a1",
      "load_balancing_algorithm": "round-robin",
      "route_policy_scope": "any"
    }
  ],
  ```

**Comment by ameowlia** [comment]
✅ After route policies are created, it shows up in the gorouter routing table...

```
  "backend.apps.identity": [
    {
      "address": "10.10.0.10:61000",
      "availability_zone": "null",
      "protocol": "http1",
      "tls": true,
      "ttl": 120,
      "tags": {
        "app_id": "07a9f703-300b-40e3-81be-2f0060f7f676",
        "app_name": "backend",
        "component": "route-emitter",
        "instance_id": "0",
        "organization_id": "f77f7566-0895-42f9-b0a7-cde95989fb1b",
        "organization_name": "o",
        "process_id": "07a9f703-300b-40e3-81be-2f0060f7f676",
        "process_instance_id": "8124e938-9e7c-4409-7732-81a1",
        "process_type": "web",
        "source_id": "07a9f703-300b-40e3-81be-2f0060f7f676",
        "space_id": "bf561092-f646-408c-ab70-3e95f597f076",
        "space_name": "s"
      },
      "private_instance_id": "8124e938-9e7c-4409-7732-81a1",
      "server_cert_domain_san": "8124e938-9e7c-4409-7732-81a1",
      "load_balancing_algorithm": "round-robin",
      "route_policy_scope": "any",
      "route_policies": [
        "cf:app:59f7aef6-9bb2-4c28-9daf-bd9bbbc1733a"
      ]
    }
    ```

ameowlia (approved):

PR #542, opened 2026-03-20
feat: add CF private IPs as route-service blocklist defaults

plowin:

Summary
---------------
* Applies the Secure-By-Default principle
* Includes private IPv4 address ranges from RFC 1918 (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16)
* Also includes the link-local range (169.254.0.0/16), see https://docs.cloudfoundry.org/concepts/asg.html#default-asg

Backward Compatibility
---------------
Breaking Change? **Yes**
* users can no longer access internal network ranges via route-services
* if operators want to keep allowing this, they need to patch this property via:
```
- type: replace
  path: /instance_groups/name=router/jobs/name=gorouter/properties/router/route_services?/egress_blocklist
  value: []
```

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -250,7 +250,11 @@ properties:
   router.route_services.egress_blocklist:
       description: |
         Route Service Requests to these CIDRs will be blocked with HTTP status code 502.
-      default: []
+      default:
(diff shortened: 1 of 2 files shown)


b1tamara (approved):
LGTM

PR #546, opened 2026-03-26
Support multiple JSESSIONID cookies for CHIPS migration

hoffmaen:

Summary
---------------
Fixes Problem 2 from #529.

During a [CHIPS migration](https://developers.google.com/privacy-sandbox/cookies/chips), an application may set **multiple session cookies with the same name** in a single response — typically a new partitioned `JSESSIONID` alongside a non-partitioned delete (`MaxAge=-1`). Previously, gorouter only translated the *first* matching session cookie into a `__VCAP_ID__` + `__VCAP_ID_META__` pair, silently dropping the second. This broke the migration because the browser never received the corresponding `__VCAP_ID__` delete cookie for the non-partitioned variant (or vice-versa).

This change makes `getSessionCookies` collect **all** matching session cookies from the response and emits a `__VCAP_ID__` + `__VCAP_ID_META__` pair for each one, preserving per-cookie attributes (`Partitioned`, `SameSite`, `MaxAge`, `Expires`, `Secure`).

Backward Compatibility
---------------
Breaking Change? **No**

AI Usage
---------------
The work was done with the assistance of Claude (claude-opus-4-6, Anthropic) via claude code.


src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -535,64 +535,63 @@ func setupStickySession(
...
+		for _, sc := range sessionCookies {
+			vcapCookie := newVcapCookie()
+			vcapCookie.Secure = vcapCookie.Secure || sc.Secure
+			vcapCookie.SameSite = sc.SameSite
+			vcapCookie.Partitioned = sc.Partitioned
+			vcapCookie.MaxAge = sc.MaxAge
+			vcapCookie.Expires = sc.Expires
+			addCookieToResponse(response, vcapCookie, logger)
+			addCookieToResponse(response, createVcapMetaCookie(vcapCookie), logger)
+		}
```
maxmoehl, line 570: Since you moved the `addCookieToResponse` call inside the loop, please get rid of the `newVcapCookie` function and just use struct literals to make this a bit more readable.
```suggestion
		for _, sc := range sessionCookies {
			vcapCookie := &http.Cookie{
				Secure: vcapCookie.Secure || sc.Secure
				// ...
			}
			addCookieToResponse(response, vcapCookie, logger)
			addCookieToResponse(response, createVcapMetaCookie(vcapCookie), logger)
		}
```

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -535,64 +535,63 @@ func setupStickySession(
...
 	} else if staleSessionScenario {
+		vcapCookie := newVcapCookie()
```
maxmoehl, line 572: Same here.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -535,64 +535,63 @@ func setupStickySession(
...
 
-// getSessionCookies returns the session cookie and __VCAP_ID__ cookie from the response, when they exist
-func getSessionCookies(response *http.Response, stickySessionCookieNames config.StringSet) (sessionCookie *http.Cookie, vcapIdCookie *http.Cookie) {
+// getSessionCookies returns all session cookies and the __VCAP_ID__ cookie from the response, when they exist.
+// Multiple session cookies may be present during a CHIPS migration (partitioned + non-partitioned delete).
+func getSessionCookies(response *http.Response, stickySessionCookieNames config.StringSet) (sessionCookies []*http.Cookie, vcapIdCookie *http.Cookie) {
 	for _, cookie := range response.Cookies() {
 		if cookie.Name == VcapCookieId {
 			vcapIdCookie = cookie
 			return
```
maxmoehl: This early return will cause bugs if the `__VCAP_ID__` cookie is the first one to occur as you will never see the session cookies then. This should probably be a if-else-if statement without any early returns.

This then raises the question what to do if there are multiple `__VCAP_ID__` cookies found? This only ever returns one.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -535,64 +535,63 @@ func setupStickySession(
...
 		}
-		if sessionCookie == nil {
-			if _, ok := stickySessionCookieNames[cookie.Name]; ok {
-				sessionCookie = cookie
-			}
+		if _, ok := stickySessionCookieNames[cookie.Name]; ok {
+			sessionCookies = append(sessionCookies, cookie)
 		}
 	}
 	return
```
maxmoehl: It has been like this before, but please add the variables to the return statement.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
hoffmaen: The early return kicks in, when the response contains the `__VCAP_ID__` cookie. It's supposed to skip sticky session creation in this case, so it also doesn't care about potential session cookies in the response. It's behaving like this, because we do not want to overwrite the `__VCAP_ID__ cookie`, when it is set by the application.

maxmoehl: Right, if there is one it returns right away.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
hoffmaen: Hm, I don't think it becomes more readable when removing the function that creates the cookies with default values.

It'd become more repetitive however. In the `staleSessionScenario`, we'd first explicitly create the struct with default values, then conditionally set most of the parameters again.

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
hoffmaen: Done

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
a18e: I'm with @hoffmaen on this one.

a18e (approved):
LGTM after a round of internal review/adjustments

PR #549, opened 2026-03-31
feat: Support __Host- prefixed session cookies for sticky sessions

hoffmaen:

Summary
---------------
 - Gorouter now recognises `__Host-`-prefixed variants of configured sticky session cookie names (e.g. `__Host-JSESSIONID`), both in application responses and client requests
  - Uses exact prefix matching (`__Host-`) per [RFC 6265bis](https://www.rfc-editor.org/rfc/draft-ietf-httpbis-rfc6265bis-19.html#name-the-__host-prefix), which mandates this canonical casing for user agents — no case-insensitive matching needed
  - No additional configuration required; the `__Host-` prefix is handled automatically for every name in `router.sticky_session_cookie_names`

Motivation
---------------
The `__Host-` cookie prefix ensures that session cookies are bound to the exact origin host and are never sent to subdomains, preventing session cookie leakage to other applications on shared domains. This prefix also enforces that cookies are only set over HTTPS, without a `Domain` attribute, and with `Path=/` — supporting it in gorouter's sticky session handling allows applications to leverage these stricter security guarantees while retaining session affinity.

Backward Compatibility
---------------
Breaking Change? **No**

Note on AI usage
---------------

### DIFF
--- a/docs/03-how-to-use-session-affinity.md
+++ b/docs/03-how-to-use-session-affinity.md
@@ -106,6 +106,21 @@ for the old non-partitioned `__VCAP_ID__` cookie alongside the new partitioned o
 
 <img src="images/sticky_sessions_chips_migration.png" alt="Sticky Sessions - CHIPS migration sequence" width="800">
 
+### Does Gorouter support `__Host-` prefixed session cookies?
+Yes. [RFC 6265bis](https://www.rfc-editor.org/rfc/draft-ietf-httpbis-rfc6265bis-19.html#name-the-__host-prefix) defines
+the `__Host-` cookie prefix, which instructs browsers to enforce additional security constraints
+(the cookie must be `Secure`, must not specify a `Domain`, and the `Path` must be `/`).
+
+Gorouter recognises cookies that use the exact `__Host-` prefix (case-sensitive, matching the
+canonical casing mandated by the RFC) in front of a configured sticky session cookie name. For
+example, if `JSESSIONID` is configured as a sticky session cookie name, Gorouter will also
+recognise `__Host-JSESSIONID` as a sticky session cookie — both in application responses (to
+create the `__VCAP_ID__` + `__VCAP_ID_META__` pair) and in client requests (to route to the
+sticky backend).
+
+No additional configuration is required; the `__Host-` prefix is handled automatically for every
+name listed in `router.sticky_session_cookie_names`.
+
 ### What happens if only one of `JSESSIONID` or `__VCAP_ID__` cookies is set on a request?
 Gorouter requires both `JSESSIONID` and `__VCAP_ID__` to be present for sticky session routing.
 If only one of them is present, Gorouter will route the request to a random available application
--- a/docs/03-how-to-use-session-affinity.md
+++ b/docs/03-how-to-use-session-affinity.md
@@ -107,7 +107,7 @@ for the old non-partitioned `__VCAP_ID__` cookie alongside the new partitioned o
 <img src="images/sticky_sessions_chips_migration.png" alt="Sticky Sessions - CHIPS migration sequence" width="800">
 
 ### Does Gorouter support `__Host-` prefixed session cookies?
-Yes. [RFC 6265bis](https://www.rfc-editor.org/rfc/draft-ietf-httpbis-rfc6265bis-19.html#name-the-__host-prefix) defines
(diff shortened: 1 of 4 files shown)


docs/03-how-to-use-session-affinity.md
```diff
@@ -106,6 +106,37 @@ for the old non-partitioned `__VCAP_ID__` cookie alongside the new partitioned o
 
 <img src="images/sticky_sessions_chips_migration.png" alt="Sticky Sessions - CHIPS migration sequence" width="800">
 
+### Does Gorouter support `__Host-` prefixed session cookies?
+Yes. [RFC 6265bis](https://www.rfc-editor.org/rfc/draft-ietf-httpbis-rfc6265bis-19.html#name-the-__host-prefix) defines
```
a18e: This link seems broken search doesn't return anything, how about https://datatracker.ietf.org/doc/draft-ietf-httpbis-rfc6265bis/ ?
(though that one will probably also change once it's no longer draft)

docs/03-how-to-use-session-affinity.md
```diff
@@ -106,6 +106,37 @@ for the old non-partitioned `__VCAP_ID__` cookie alongside the new partitioned o
...
+
+In practice this is not a concern: unlike CHIPS migration, there is no need to set both cookies in
+the same response. Because `JSESSIONID` and `__Host-JSESSIONID` are distinct cookie names in the
+browser's jar, the expected migration path is for the application to simply stop setting
+`JSESSIONID` and start setting `__Host-JSESSIONID` — the old cookie expires naturally.
+
+Note: if an application were to set a new `__Host-JSESSIONID` alongside a delete (`Max-Age=0`) for
+the old `JSESSIONID` in the same response, both would produce a `__VCAP_ID__` in the same cookie
+jar partition. Depending on processing order, the browser could apply the delete `__VCAP_ID__`
+after the new one, effectively removing it.
```
a18e: Maybe add what this actually means for a developer, e.g. "This could mean breaking session stickyness for the time of migration from JSESSIONID to _Host-...?"

src/code.cloudfoundry.org/gorouter/handlers/helpers.go
```diff
@@ -87,9 +87,15 @@ func GetStickySession(request *http.Request, stickySessionCookieNames config.Str
...
-	for stickyCookieName := range stickySessionCookieNames {
-		if _, err := request.Cookie(stickyCookieName); err == nil {
+
+	// Try choosing a backend using sticky session.
+	// Also match the "__Host-" prefixed variant of each configured cookie name (RFC 6265bis).
+	for _, c := range request.Cookies() {
+		name := c.Name
+		if strings.HasPrefix(name, "__Host-") {
+			name = name[7:]
+		}
```
a18e:
```suggestion
		name := strings.TrimPrefix(c.Name, "__Host-")
```
Wouldn't this be simpler? see also below

src/code.cloudfoundry.org/gorouter/proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -597,13 +597,26 @@ func getSessionCookies(response *http.Response, stickySessionCookieNames config.
...
+	if _, ok := stickySessionCookieNames[cookieName]; ok {
+		return true
+	}
+	if strings.HasPrefix(cookieName, "__Host-") {
+		_, ok := stickySessionCookieNames[cookieName[7:]]
+		return ok
+	}
+	return false
+}
+
```
a18e, line 614:
```suggestion
		// also match __Host- prefix
		name := strings.TrimPrefix(cookie.Name, "__Host-")
		if _, ok := stickySessionCookieNames[name]; ok {

			sessionCookies = append(sessionCookies, cookie)
		}
	}
	return sessionCookies, nil
}
```

hoffmaen: I simplified it to use `strings.TrimPrefix` but decided to keep `IsSessionCookie` as a separate exported function — it has dedicated unit tests that cover edge cases (wrong casing, missing hyphen, unknown names), which I find easier to maintain as a standalone helper.

a18e (approved):
LGTM

PR #550, opened 2026-04-08
feat: Add endpoints_per_pool Prometheus metric for hash-based routing

hoffmaen:

Linked with https://github.com/cloudfoundry/docs-running-cf/pull/143

Summary
---------------
- Introduces a new `endpoints_per_pool` Prometheus gauge metric with labels `route` and `lb_algorithm`, reporting the number of endpoints per route pool for routes using hash-based load balancing
- The metric is updated on endpoint registration, unregistration, and pruning of stale endpoints
- When a route switches away from hash-based routing (including removal of the HB setting), the metric entry is automatically cleaned up

Motivation
---------------
Hash-based routing distributes requests based on a consistent hash, which can degrade when the instance count for a pool becomes too high. This metric provides visibility into the endpoint count per hash-based routing pool, keeping metric cardinality low by only tracking hash-based routes.

Backward Compatibility
---------------
Breaking Change? **No**

Note on AI usage
---------------

Co-authored-by: Soha-Albaghdady <soha.alboghdady@sap.com>

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/metrics/compositereporter.go
+++ b/src/code.cloudfoundry.org/gorouter/metrics/compositereporter.go
@@ -48,7 +48,7 @@ type MetricReporter interface {
 	CaptureNATSDroppedMessages(messages int)
 	UnmuzzleRouteRegistrationLatency()
 	CaptureEndpointsPerPool(endpoints int, route string, lbAlgo string)
-	DeleteEndpointsPerPool(route, lbAlgo string)
+	UncaptureEndpointsPerPool(route, lbAlgo string)
 }
 
 type ComponentTagged interface {
@@ -238,9 +238,9 @@ func (m MultiMetricReporter) CaptureEndpointsPerPool(endpoints int, route string
 	}
 }
 
-func (m MultiMetricReporter) DeleteEndpointsPerPool(route, lbAlgo string) {
+func (m MultiMetricReporter) UncaptureEndpointsPerPool(route, lbAlgo string) {
 	for _, r := range m {
-		r.DeleteEndpointsPerPool(route, lbAlgo)
+		r.UncaptureEndpointsPerPool(route, lbAlgo)
 	}
 }
 
@@ -276,6 +276,6 @@ func (c *CompositeReporter) CaptureEndpointsPerPool(endpoints int, route string,
 	c.MetricReporter.CaptureEndpointsPerPool(endpoints, route, lbAlgo)
 }
 
-func (c *CompositeReporter) DeleteEndpointsPerPool(route, lbAlgo string) {
-	c.MetricReporter.DeleteEndpointsPerPool(route, lbAlgo)
+func (c *CompositeReporter) UncaptureEndpointsPerPool(route, lbAlgo string) {
+	c.MetricReporter.UncaptureEndpointsPerPool(route, lbAlgo)
 }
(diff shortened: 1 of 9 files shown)


src/code.cloudfoundry.org/gorouter/metrics/compositereporter.go
```diff
@@ -257,3 +271,11 @@ func (c *CompositeReporter) CaptureRoutingResponseLatency(b *route.Endpoint, sta
 func (c *CompositeReporter) CaptureHTTPLatency(d time.Duration, sourceID string) {
 	c.MetricReporter.CaptureHTTPLatency(d, sourceID)
 }
+
+func (c *CompositeReporter) CaptureEndpointsPerPool(endpoints int, route string, lbAlgo string) {
+	c.MetricReporter.CaptureEndpointsPerPool(endpoints, route, lbAlgo)
+}
+
+func (c *CompositeReporter) DeleteEndpointsPerPool(route, lbAlgo string) {
```
b1tamara: I don't like the function name. Maybe "UncaptureEndpointsPerPool"? It is not always a deletion case; it can also be a change to the load-balancing algorithm in a pool.

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -461,6 +474,22 @@ func (r *RouteRegistry) freshenRoutes() {
...
+			return
+		}
+		r.reporter.CaptureEndpointsPerPool(pool.NumEndpoints(), string(uri), config.LOAD_BALANCE_HB)
+	} else {
+		// Delete stale entries for routes that have switched from HB to another load balancing algorithm.
+		// This is called unconditionally because the pool may still report its algorithm as HB
+		// even after the HB setting was removed (setPoolLoadBalancingAlgorithm ignores empty values).
+		r.reporter.DeleteEndpointsPerPool(string(uri), config.LOAD_BALANCE_HB)
+	}
+}
```
b1tamara, line 494: For better readability:
```
func (r *RouteRegistry) reportEndpointsPerPool(uri route.Uri, endpoint *route.Endpoint) {
	if endpoint.LoadBalancingAlgorithm != config.LOAD_BALANCE_HB {
		// Delete stale entries for routes that have switched from HB to another load balancing algorithm.
		// This is called unconditionally because the pool may still report its algorithm as HB
		// even after the HB setting was removed (setPoolLoadBalancingAlgorithm ignores empty values).
		r.reporter.DeleteEndpointsPerPool(string(uri), config.LOAD_BALANCE_HB)
		return
	}

	r.RLock()
	pool := r.byURI.Find(uri.RouteKey())
	r.RUnlock()

	if pool == nil || pool.NumEndpoints() == 0 {
		r.reporter.DeleteEndpointsPerPool(string(uri), config.LOAD_BALANCE_HB)
		return
	}
	r.reporter.CaptureEndpointsPerPool(pool.NumEndpoints(), string(uri), config.LOAD_BALANCE_HB)
}
```

src/code.cloudfoundry.org/gorouter/registry/registry.go
```diff
@@ -461,6 +474,22 @@ func (r *RouteRegistry) freshenRoutes() {
 	})
 }
 
+func (r *RouteRegistry) reportEndpointsPerPool(uri route.Uri, endpoint *route.Endpoint) {
```
b1tamara, line 479: Could you add a function comment to make more clear that it reports only for hash-based routing route pools?

src/code.cloudfoundry.org/gorouter/metrics/compositereporter.go
hoffmaen: Done

src/code.cloudfoundry.org/gorouter/registry/registry.go
hoffmaen: Done.

src/code.cloudfoundry.org/gorouter/registry/registry.go
hoffmaen: Done.

b1tamara (approved):

PR #552, opened 2026-04-09
fix: Per-route load balancing algorithm not reverting to platform default on removal

### LINKED ISSUE #551: Per-route load balancing algorithm is not reset to platform default when removed from a route
Current behavior
----------

When a per-route load balancing algorithm is configured (e.g., `hash` for hash-based routing) and later removed via `cf update-route -r loadbalancing`, gorouter's `setPoolLoadBalancingAlgorithm` ignores the empty value and the `EndpointPool` retains the previously set algorithm. This means the route continues to use the per-route algorithm (e.g., hash-based routing with its Maglev lookup table) instead of reverting to the platform default (e.g., `round-robin`).

The root cause is in `mbus/subscriber.go`: when the `loadbalancing` field is absent or empty in the NATS message, `makeEndpoint` leaves the endpoint's `LoadBalancingAlgorithm` as an empty string. In `route/pool.go`, `setPoolLoadBalancingAlgorithm` treats an empty string as "no change", so the pool keeps the old algorithm.

Desired behavior
----------

When the `loadbalancing` field is absent or empty in the NATS route registration message, gorouter should default the endpoint's load balancing algorithm to the platform default configured via `router.balancing_algorithm`. This ensures that removing a per-route algorithm causes the pool to revert to the platform default. Hash-based routing state (`HashLookupTable`, `HashRoutingProperties`) should be cleaned up when switching away from hash-based routing.

Affected Version
----------

develop

hoffmaen:

Summary
---------------
Fixes the bug where removing a per-route load balancing algorithm (e.g., via `cf update-route -r loadbalancing`) left the pool using the old algorithm instead of reverting to the platform default. Fixes #551                                                               
   
Changes
---------------

- Default the endpoint's load balancing algorithm to the global platform algorithm (`config.LoadBalance`) in `mbus/subscriber.go`; only override when the NATS message contains a non-empty value                                                                                                                                                        
- Clean up hash-based routing state (`HashLookupTable`, `HashRoutingProperties`) in `EndpointPool.setPoolLoadBalancingAlgorithm` when switching away from hash-based routing                                                                                        
                                                                                                                                                                                                                                                                                   

Backward Compatibility
---------------
Breaking Change? **No**

Note on AI usage
---------------

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/route/pool.go
+++ b/src/code.cloudfoundry.org/gorouter/route/pool.go
@@ -207,10 +207,9 @@ type EndpointPool struct {
 	random                 *rand.Rand
 	logger                 *slog.Logger
 	updatedAt              time.Time
-	LoadBalancingAlgorithm        string
-	defaultLoadBalancingAlgorithm string
-	HashRoutingProperties         *HashRoutingProperties
-	HashLookupTable               MaglevLookup
+	LoadBalancingAlgorithm string
+	HashRoutingProperties  *HashRoutingProperties
+	HashLookupTable        MaglevLookup
 }
 
 type EndpointOpts struct {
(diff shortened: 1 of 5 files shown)


src/code.cloudfoundry.org/gorouter/route/pool.go
```diff
@@ -667,18 +667,25 @@ func (p *EndpointPool) MarshalJSON() ([]byte, error) {
 }
 
 // setPoolLoadBalancingAlgorithm overwrites the load balancing algorithm of a pool by that of a specified endpoint, if that is valid.
+// An empty algorithm is treated as "no change" — the pool keeps its current algorithm.
```
b1tamara: Is it possible at all that `endpoint.LoadBalancingAlgorithm == ""` will be true? Or is it the case that the lb was never set via per-route option?

b1tamara: Looking at this [line](https://github.com/cloudfoundry/routing-release/pull/552/changes#diff-a3cb8cde907ca8ef9a573547cc9464f3bfa9348ce3beedcb838e768d40a99c85R85), I think we always set the algo (either globalRoutingAlgo or configured per-route) for every endpoint registration. At the later time point in the `pool.Put` function, it will be decided if it is a new endpoint, an update, or only a refresh

hoffmaen: Good catch — removed the guard and its tests in since it's unreachable with the plain string type.

b1tamara (dismissed):
LGTM

b1tamara (approved):

PR #554, opened 2026-04-14
fix: Handle empty LoadBalancingAlgorithm in pool and update subscriber test expectations

hoffmaen:

Summary
---------------
- Re-added the empty-string guard in `setPoolLoadBalancingAlgorithm` to avoid spurious error logs caused by test endpoints that don't set a `LoadBalancingAlgorithm`
- Updated subscriber test expectations to account for `makeEndpoint` now defaulting to the global routing algorithm (`round-robin`) when none is provided in the NATS message

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
+++ b/src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
@@ -372,10 +372,11 @@ var _ = Describe("Subscriber", func() {
 			Eventually(registry.RegisterCallCount).Should(Equal(1))
 			_, originalEndpoint := registry.RegisterArgsForCall(0)
 			expectedEndpoint := route.NewEndpoint(&route.EndpointOpts{
-				Host:             "host",
-				AppId:            "app",
-				Protocol:         "http1",
-				AvailabilityZone: "zone-meow",
+				Host:                   "host",
+				AppId:                  "app",
+				Protocol:               "http1",
+				AvailabilityZone:       "zone-meow",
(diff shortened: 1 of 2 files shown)


src/code.cloudfoundry.org/gorouter/mbus/subscriber_test.go
```diff
@@ -436,9 +438,10 @@ var _ = Describe("Subscriber", func() {
 			Eventually(registry.RegisterCallCount).Should(Equal(1))
 			_, originalEndpoint := registry.RegisterArgsForCall(0)
 			expectedEndpoint := route.NewEndpoint(&route.EndpointOpts{
-				Host:     "host",
-				AppId:    "app",
-				Protocol: "http1",
+				Host:                   "host",
+				AppId:                  "app",
+				Protocol:               "http1",
+				LoadBalancingAlgorithm: config.LOAD_BALANCE_RR,
```
b1tamara: Only one note to all tests. Prefer to use the cfg.LoadBalance (set in DefaultConfig) instead of hard-coded LOAD_BALANCE_RR:
```
				Host:                   "host",
				AppId:                  "app",
				Protocol:               "http1",
				LoadBalancingAlgorithm: cfg.LoadBalance,
			})
```

hoffmaen: Good catch, I changed it to use the default from the config.

b1tamara (approved):
LGTM

PR #555, opened 2026-04-14
Fix flaky tests

hoffmaen:

Summary
---------------
Fix multiple sources of test flakiness across the gorouter test suites (router, integration, registry). These fixes address intermittent failures observed under CI load with 7 parallel Ginkgo nodes.

Changes
---------------
- **Fix data race in registry pruning test** — Background goroutine keeping an endpoint alive could outlive its spec, racing with the next spec's `BeforeEach` overwriting the shared `configObj`. Fixed with a `stoppedChan` pattern to drain the goroutine before the spec exits.                                                                                                                                                                                                    
- **Set `SetDefaultEventuallyTimeout(10s)` globally** in `router_suite_test.go` and `integration/init_test.go`. Both suites relied on Gomega's 1s default, which is too short under CI load.
- **Add `Timeout: 500ms` to health check HTTP client** in `testHealthCheckEndpoint` — An `http.Client{}` with no timeout can hang indefinitely, causing Ginkgo to report "timed out waiting for all parallel procs".                                                                                                                                                                                                                                                                   
- **Replace `Eventually(...).ShouldNot` with `Consistently(...).ShouldNot`** in `health_listener_test.go` and `integration/main_test.go` — `Eventually(...).ShouldNot(...)` waits the full timeout (now 10s) before passing, while `Consistently` is the semantically correct choice for "this should never happen" assertions.                                                                                                                                                        
- **Fix flaky keep-alive and 100-Continue tests** — Increase `EndpointTimeout` from 500ms to 2s in the HTTP keep-alive context (only 125ms margin was left for request processing), and add a 10s read deadline on the raw TCP connection in the 100-Continue test.

Backward Compatibility
---------------
Breaking Change? **No**
<!---
If this is a breaking change, or modifies currently expected behaviors of core functionality

- Has the change been mitigated to be backwards compatible?
- Should this feature be considered experimental for a period of time, and allow operators to opt-in?
- Should this apply immediately to all deployments?
-->

Note on AI usage
---------------

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/integration/init_test.go
+++ b/src/code.cloudfoundry.org/gorouter/integration/init_test.go
@@ -48,7 +48,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	json.Unmarshal(data, &res)
 	gorouterPath = res.Gorouter
 	testAssets = res.Test
-	SetDefaultEventuallyTimeout(5 * time.Second)
+	SetDefaultEventuallyTimeout(10 * time.Second)
 	SetDefaultEventuallyPollingInterval(100 * time.Millisecond)
 	SetDefaultConsistentlyDuration(1 * time.Second)
 	SetDefaultConsistentlyPollingInterval(10 * time.Millisecond)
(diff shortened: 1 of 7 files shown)


plowin (dismissed):
LGTM!

kart2bc (dismissed):

plowin (dismissed):
Check if this re-triggers validation

ameowlia (approved):

PR #561, opened 2026-05-04
Fixing bug of missing access logs

### LINKED ISSUE #564: Fix missing Gorouter access logs when a client disconnects mid-stream.
### Current behavior

Recently, we encountered an issue in Gorouter where a some access logs are missing when a client sends RST during streaming.

**In details :** 
When a client aborted a TCP connection while receiving a response, the access log entry for that request was silently dropped. This happened because the client disconnect triggered an internal panic that interrupted the request handling chain before the logging step could execute, leaving no trace of the request in the access log.

### Desired behavior

Every request now produces an access log entry, regardless of whether the client disconnects mid-stream. 

### Note on AI usage
Parts of this code and tests were developed with assistance from Claude Code (claude-sonnet-4-6).

### Affected Version

0.378

brmlad:

Summary
---------------
We propose introducing new logic in Gorouter so every request now produces an access log entry, regardless of whether the client disconnects mid-stream. The fix ensures the logging step always runs, even when a client abort occurs during response streaming. Requests affected by a client disconnect are recorded with the state captured at the time of the abort, including status code, bytes sent, and timings, and are marked with an error indicating the client closed the connection during streaming.

Resolves: https://github.com/cloudfoundry/routing-release/issues/564
---------------

Backward Compatibility
---------------
Breaking Change? **No**

Note on AI usage
---------------
Parts of this code and tests were developed with assistance from Claude Code (claude-sonnet-4-6).

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/handlers/access_log_test.go
+++ b/src/code.cloudfoundry.org/gorouter/handlers/access_log_test.go
@@ -2,6 +2,7 @@ package handlers_test
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -70,7 +71,7 @@ var _ = Describe("AccessLog", func() {
 	})
 
 	testProxyWriterHandler := func(rw http.ResponseWriter, req *http.Request, next http.HandlerFunc) {
-		proxyWriter := utils.NewProxyResponseWriter(rw)
+		proxyWriter := utils.NewProxyResponseWriter(rw, logger.Logger)
 		next(proxyWriter, req)
 	}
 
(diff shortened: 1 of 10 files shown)


src/code.cloudfoundry.org/gorouter/handlers/access_log_test.go
```diff
@@ -196,4 +197,36 @@ var _ = Describe("AccessLog", func() {
 		})
 	})
 
+	Context("when the client disconnects during response streaming", func() {
```
ameowlia, line 200: The new test only covers one branch of this condition:

```
if alr.RouterError == "" && proxyWriter.WriteError() != nil {
    alr.RouterError = utils.ConnectionCloseDuringStreamingErrMsg
}
```

Could you also add a test for the case where both an existing RouterError is set on the response and a write error occurred?

ameowlia (changes requested):

src/code.cloudfoundry.org/gorouter/handlers/access_log_test.go
b1tamara: Implemented a new test when RouterError exists on the response header, and we do not want to overwrite it.

ameowlia (approved):

PR #537, opened 2026-03-09
Egress Block List for Gorouter Route Services

Dariquest:
Summary
---------------
**Egress Blocklist for Route Services**

Operators will be enabled to provide a list of IP addresses, which can't be accessed via route services.
The list of blocked CIDRs for route services is configured in the Gorouter configuration property **router.route_services.block_CIDRs**. Route Service Requests to these CIDRs will be blocked with HTTP status code 502.

Custom Route Service Dialer enables callback to make decisions after DNS resolution and before the connection is established by providing a Control function. This function is used to abort the Dialer's connection to a particular target before it is opened based on the defined block list. 

Backward Compatibility
---------------
Breaking Change? **No**

PR #572, opened 2026-07-03
Egress Block-Listing for Gorouter Route Services: Add IP Normalisation

Dariquest:
Summary
---------------
Normalise an IP address for block-listing introduced in [Egress Block List for Gorouter Route Services
](https://github.com/cloudfoundry/routing-release/pull/537) to eliminate two documented blind spots of netip.Prefix.Contains():

* IPv4-mapped IPv6 addresses (::ffff:192.168.1.1) are not matched against an IPv4 prefix (192.168.1.0/24) — Go explicitly documents this behaviour.
* IPv6 addresses with a zone identifier (fe80::1%lo0) are never matched by any prefix — Contains always returns false for zoned addresses.

The fix normalises the address before the block-list check:

addr := addrPort.Addr().WithZone("").Unmap()

.WithZone("") — strips the zone identifier (%lo0), so fe80::1%lo0 → fe80::1.
.Unmap() — converts an IPv4-mapped IPv6 address to its plain IPv4 form, so ::ffff:192.168.1.1 → 192.168.1.1.

After normalisation, Prefix.Contains() behaves correctly for both cases.

Backward Compatibility
---------------
Breaking Change? **No**

Note on AI usage
---------------
Parts of this code and tests were developed with assistance of GitHub Copilot (Model: claude-sonnet-4-6).

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/proxy/proxy.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/proxy.go
@@ -336,9 +336,14 @@ func RouteServiceDialControl(routeServiceConfig *routeservice.RouteServiceConfig
 			return fmt.Errorf("wrong address format '%s': an IP and a port number are expected", address)
 		}
 
+		// Normalize the address: strip any IPv6 zone identifier (e.g. fe80::1%lo0)
+		// and unmap IPv4-mapped IPv6 addresses (e.g. ::ffff:192.168.1.1 → 192.168.1.1)
+		// so that prefix matching works correctly in both cases.
+		addr := addrPort.Addr().WithZone("").Unmap()
+
 		for _, blockedIP := range routeServiceConfig.EgressBlockList() {
-			if blockedIP.Contains(addrPort.Addr()) {
-				return fmt.Errorf("connection to %s not allowed", addrPort.Addr().String())
+			if blockedIP.Contains(addr) {
+				return fmt.Errorf("connection to %s not allowed", addr.String())
 			}
 		}
 		return nil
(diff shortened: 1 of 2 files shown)


geofffranks:
Thanks! Claude had a couple minor concerns:

1. The error message now reports the normalized address rather than the raw one the caller presented:
```
return fmt.Errorf("connection to %s not allowed", addr.String())
// addr = "192.168.1.1", but caller connected to "::ffff:192.168.1.1"
An operator debugging a blocked connection may be confused that the error says 192.168.1.1 when the actual dialed address was ::ffff:192.168.1.1. Consider logging or surfacing the original address, e.g.:
return fmt.Errorf("connection to %s not allowed", addrPort.Addr().String())
// Or: "connection to %s (normalized: %s) not allowed", addrPort.Addr(), addr
The current behavior is still correct and safe — this is a quality-of-life point, not a bug.
```
2.  Pure IPv6 blocked by IPv6 prefix — There's still no positive test for a plain (no zone, no mapping) IPv6 address being blocked. (pre-existing)
3. Zone-stripped non-match — No test verifies that an address with a zone that falls outside all blocklist prefixes is still allowed after normalization.

It seems like we'd probably want at least 1 and 3 fixed, right?

Dariquest:
> 3\. verifies that an address with a zone that falls outside all blocklist prefixes is still allowed after normalization.

Thanks @geofffranks:
1. Added the error message
2. Added a test of a plain IPv6 address (no zone, no IPv4-mapping) being correctly blocked when it falls inside an IPv6 prefix in the blocklist. This closes the gap identified in the Claude review — there was no positive test proving that a straightforward IPv6 address like 2001:db8::1 gets blocked by a matching IPv6 prefix 2001:db8::/32.
3. Added a test "allows connection when IPv6 address with zone falls outside all blocklist prefixes", which verifies that fe80::1%lo0 with a zone is still allowed when the prefix list doesn't contain it, ensuring zone-stripping doesn't cause false positives.

geofffranks (approved):

PR #581, opened 2026-08-17
Preserve all X-Forwarded-For header lines

### LINKED ISSUE #580: X-Forwarded-For loses entries when the header arrives as multiple lines
### Current behavior

Since v0.386.0, gorouter drops all but the first `X-Forwarded-For` value when a request
arrives with the header split across multiple header lines.

In #573 the reverse proxy was migrated from `Director` to `Rewrite` mode (`Director` is
deprecated in newer Go and was flagged by staticcheck). In `Rewrite`
mode `httputil.ReverseProxy` strips the `X-Forwarded-*` headers from the outgoing request,
so `proxy.go` rebuilds `X-Forwarded-For` by hand:

```go
if prior := r.In.Header.Get("X-Forwarded-For"); prior != "" {
    r.Out.Header.Set("X-Forwarded-For", prior+", "+clientIP)
}
```

`Header.Get` only returns the *first* `X-Forwarded-For` header line. A request can
legitimately carry several — for example when a proxy in front of gorouter (in our case
HAProxy with `option forwardfor`) appends its own `X-Forwarded-For` line in addition to
one the client or an upstream hop already set. Every line after the first is silently
dropped, and the appended IP is often the one that mattered (the true client IP as seen
by the fronting proxy).

Example, same request before and after the upgrade:

```
before (v0.385.0): x-forwarded-for: 3.125.117.248, 52.58.41.143, 10.0.72.1
after  (v0.386.0+): x-forwarded-for: 3.125.117.248, 10.0.72.1
```

The `52.58.41.143` entry (added as a second header line) is gone. The old `Director`-based
code did not have this problem because `ReverseProxy` handled `X-Forwarded-For` itself and
folded all the inbound lines together.

### Desired behavior

All inbound `X-Forwarded-For` header lines should be preserved, matching the pre-0.386.0
behavior. Joining `r.In.Header.Values("X-Forwarded-For")` instead of using `Header.Get`
restores this.

### Affected Version

v0.386.0 through current `develop` (introduced in #573, unchanged as of v0.390.0).

hoffmaen:
Fixes #580

Summary
---------------
Restores handling of `X-Forwarded-For` when it arrives as multiple header lines by joining `Header.Values("X-Forwarded-For")` instead of reading only the first line via `Header.Get`, a regression introduced with the `Director`→`Rewrite` migration in #573. Fixes #580.

Backward Compatibility
---------------
Breaking Change? **No**

Note on AI Usage
---------------
Parts of this code and tests were developed with assistance from Claude Code (claude-opus-4.8).

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/proxy/proxy.go
+++ b/src/code.cloudfoundry.org/gorouter/proxy/proxy.go
@@ -154,7 +154,12 @@ func NewProxy(
 				r.Out.Header.Set("X-Forwarded-Host", host)
 			}
 			if clientIP, _, err := net.SplitHostPort(r.In.RemoteAddr); err == nil {
-				if prior := r.In.Header.Get("X-Forwarded-For"); prior != "" {
+				// Join ALL inbound X-Forwarded-For header lines, not just the
+				// first. A request can legitimately arrive with multiple
+				// X-Forwarded-For header lines (e.g. a proxy in front of gorouter
+				// appends its own line in addition to any the client/upstream
+				// already set).
+				if prior := strings.Join(r.In.Header.Values("X-Forwarded-For"), ", "); prior != "" {
 					r.Out.Header.Set("X-Forwarded-For", prior+", "+clientIP)
 				} else {
 					r.Out.Header.Set("X-Forwarded-For", clientIP)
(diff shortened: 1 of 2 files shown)


geofffranks (approved):
Correct fix, LGTM.

`Header.Get` → `strings.Join(Header.Values(…), ", ")` is the standard Go idiom for folding multi-line headers and directly addresses the regression from the Director→Rewrite migration in #573. No behavioral change for the single-header case. The new test cleanly reproduces the bug.

Minor note (not a blocker): `X-Forwarded-Proto` and `X-Forwarded-Host` at proxy.go:150,153 use the same `Header.Get` pattern. Less likely to arrive as multiple lines in practice, but worth a follow-up issue if that's plausible in your deployment topology.

PR #584, opened 2026-09-11
fix(gorouter): hash_balance as string (partial revert of #535)

### LINKED ISSUE #585: gorouter drops hash-based route registrations — hash_balance no longer parsed when sent as a JSON string (regression introduced in v0.386.0 / #535)
### Current behavior

Since v0.386.0, gorouter fails to unmarshal `router.register` NATS messages for any route configured with hash-based load balancing (`loadbalancing: hash`) that includes a `hash_balance` factor. The message is rejected as a validation error and dropped, so the endpoint is never registered — the route becomes unreachable (404 / stale until TTL expiry) even though the app and the registration message are healthy.

**Observed log (`gorouter.subscriber`):**

```json
{"message":"validation-error","source":"gorouter.subscriber","data":{
  "error":"json: cannot unmarshal string into Go struct field RegistryMessageOpts.options.hash_balance of type float64",
  "payload":"{\"host\":\"10.0.73.2\",\"port\":0,\"tls_port\":61000,\"uris\":[\"...cf...sapcloud.io\"],\"protocol\":\"http1\",\"app\":\"e2483c52-...\",...,\"options\":{\"loadbalancing\":\"hash\",\"hash_header\":\"http-header-name\",\"hash_balance\":\"1.2\"}}",
  "subject":"router.register"}}
```

### Desired behavior

gorouter should again accept `hash_balance` when it arrives as a JSON string, matching the wire format that Cloud Controller and route-emitter actually produce.

Fix WIP: https://github.com/cloudfoundry/routing-release/pull/584

### Affected Version

v0.386.0 through v0.393.0 (latest). Introduced by commit 02a2a9353 (PR #535); last unaffected release: v0.385.0.

a18e:

Summary
---------------
This PR restores the original parsing of `hash_balance`.
The `,string` suffix was removed as part of https://github.com/cloudfoundry/routing-release/pull/535/changes#diff-a3cb8cde907ca8ef9a573547cc9464f3bfa9348ce3beedcb838e768d40a99c85L46, which broke hash-based routing.

Backward Compatibility
---------------
Breaking Change? No, fixes (unintended?) breaking change.

See https://github.com/cloudfoundry/routing-release/issues/585

### DIFF
--- a/src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
+++ b/src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
@@ -43,7 +43,7 @@ type RegistryMessage struct {
 type RegistryMessageOpts struct {
 	LoadBalancingAlgorithm string  `json:"loadbalancing"`
 	HashHeaderName         string  `json:"hash_header"`
-	HashBalance            float64 `json:"hash_balance"`
+	HashBalance            float64 `json:"hash_balance,string"`
 	// RFC route policy options (from Cloud Controller via Diego sync)
 	RoutePolicyScope   string `json:"route_policy_scope,omitempty"`
 	RoutePolicySources string `json:"route_policy_sources,omitempty"`


hoffmaen (dismissed):

src/code.cloudfoundry.org/gorouter/mbus/subscriber.go
```diff
@@ -43,7 +43,7 @@ type RegistryMessage struct {
 type RegistryMessageOpts struct {
 	LoadBalancingAlgorithm string  `json:"loadbalancing"`
 	HashHeaderName         string  `json:"hash_header"`
-	HashBalance            float64 `json:"hash_balance"`
```
b1tamara, line 46: Could you please implement a new unit test in subscriber_test.go to ensure this regression is avoided during NATS message registration.

Something like:
```
It("parses hash_balance from the incoming JSON string", func() {
				data := []byte(`{"host":"host","app":"app","protocol":"http2","uris":["test.example.com"],"options":{"loadbalancing":"hash","hash_header":"X-Header","hash_balance":"1.5"}}`)

				err := natsClient.Publish("router.register", data)
				Expect(err).ToNot(HaveOccurred())

				Eventually(registry.RegisterCallCount).Should(Equal(1))
					Host:                   "host",
					AppId:                  "app",
					Protocol:               "http2",
					LoadBalancingAlgorithm: "hash",
					HashHeaderName:         "X-Header",
					HashBalanceFactor:      1.5,
				})

				Expect(originalEndpoint).To(Equal(expectedEndpoint))
			})
```

PR #589, opened 2026-09-15
Fix RATS test by trimming orgGuid

hoffmaen:

Summary
---------------
RATS tests fail because the `orgId` returned by CF CLI has a trailing newline character. This change trims it, verified with local test execution.

Backward Compatibility
---------------
Breaking Change? **No**


b1tamara (approved):
LGTM

